PR #53, opened 2024-02-14
Updated github workflows to do manual triggers from web UI for both test and prod

binaryaaron:
* Changed to trigger releases manually on the github web UI
*  parses Ruff's version directly
* uses UV as a python package installer

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -11,6 +11,8 @@ on:
 jobs:
   ruff:
     runs-on: ubuntu-latest
+    env:
+     VENV_PATH: "$HOME/venv" 
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yaml
```diff
@@ -19,12 +19,22 @@ jobs:
         with:
           python-version: "3.10"
 
-      - name: Install Ruff
-        run: pipx install ruff==0.2.1 # Needs to match pyproject.toml
+      - name: get ruff version
```
sfriedowitz, line 22: 👍

.github/workflows/publish.yaml
```diff
@@ -5,14 +5,17 @@ on:
     tags:
       - "v*"
   pull_request:
+    types:
+        - labeled
```
sfriedowitz, line 9: Does the contributing guide need to be updated to reflect this now?

sfriedowitz (approved):
I think this is great as a first step, and much cleaner to get us up and running with the publishing workflow! Thanks for the changes.

.github/workflows/main.yaml
```diff
@@ -19,59 +19,109 @@ jobs:
...
-      - name: Lock Poetry
-        run: poetry lock
-
       - name: Set up Python 3.10
         uses: actions/setup-python@v5
         with:
           python-version: "3.10"
-          cache: poetry
+
+      - name: Install UV, venv
```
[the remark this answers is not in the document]
binaryaaron: Agreed! 

outside of this PR, yeah - we can consolidate the jobs into steps in a single job (easy), or eventually use our own image. will follow up elsewhere.

PR #54, opened 2024-02-16
RD2024-124: Job runner method

sfriedowitz:
The motivation for this PR is to greater facilitate usage of `lm-buddy` as a standalone tool outside of Ray.

A user may be interested in running an LM Buddy job from a Jupyter notebook on a dev pod entirely in-memory. This is currently possible, but a user has to manually import the correct entrypoint function from the `.jobs` module, which is intended to be a bit more shielded.

This PR introduces a generic `run_job` method that accepts any config and dispatches to the appropriate entrypoint. This seems like a cleaner public API for calling the package.

I've also gone and removed the optional dependency groups, since everything relies on the HuggingFace libraries atm (even lm-harness), and this allows us to get rid of the internal function imports in the CLI.


src/lm_buddy/jobs/common.py
```diff
@@ -0,0 +1,49 @@
...
+class LMBuddyJobConfig(BaseLMBuddyConfig):
+    """Configuration that comprises the entire input to an LM Buddy job.
+
+    This class implements helper methods for de/serializing the configuration from file.
+
+    Currently, there is a 1:1 mapping between job entrypoints and job config implementations,
+    but this is not rigidly constrained by the interface. This may change in the future.
+    """
+
+    @classmethod
```
sfriedowitz, line 28: I've intentionally moved the YAML file methods onto this subclass to restrict the interface.

Currently, the only supported case we have for writing to/from file is for full job configs, not the independent sub-components. So it feels a bit more natural to constrain that functionality to this subclass rather than on the parent `BaseLMBuddyConfig` class.

src/lm_buddy/cli/schema.py
```diff
@@ -1,5 +1,11 @@
+import json
+
 import click
 
+from lm_buddy.jobs.finetuning import FinetuningJobConfig
```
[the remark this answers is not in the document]
sfriedowitz: So the folder layout I'm thinking of is as follows:

```
lm_buddy.jobs
   configs/
   _entrypoints
   __init__.py
```

Then, the individual entrypoint functions could be better shielded from the public API since we only expose the generic `run_job` method.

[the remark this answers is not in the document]
sfriedowitz: I think ill do this re-organization in a follow up PR to avoid the clutter in this one.

src/lm_buddy/cli/run.py
```diff
@@ -11,27 +18,19 @@ def group():
 @group.command("simple", help="Run the simple test job.")
 @click.option("--config", type=str)
 def run_simple(config: str) -> None:
```
veekaybee, line 20: this is a nice refactor to include imports at the top-level, makes it much cleaer what's being run

src/lm_buddy/cli/run.py
```diff
@@ -1,6 +1,13 @@
 import click
 
-from lm_buddy.integrations.wandb import WandbArtifactLoader
+import lm_buddy
```
veekaybee, line 3: could we import run_job directly here instead of the package?

src/lm_buddy/cli/schema.py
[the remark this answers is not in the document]
veekaybee: let's discuss in that second PR, but it would be helpful to see the new workflow too

src/lm_buddy/jobs/__init__.py
```diff
@@ -0,0 +1,30 @@
+from lm_buddy.integrations.wandb import ArtifactLoader, WandbArtifactLoader
+from lm_buddy.jobs.common import LMBuddyJobConfig
+from lm_buddy.jobs.finetuning import FinetuningJobConfig, run_finetuning
+from lm_buddy.jobs.lm_harness import LMHarnessJobConfig, run_lm_harness
+from lm_buddy.jobs.simple import SimpleJobConfig, run_simple
+
+
+def run_job(
+    config: LMBuddyJobConfig,
+    *,
```
veekaybee, line 10: do we always know we're only getting config and artifact loader here? if that's the case, we can consider dropping the args for clarity

src/lm_buddy/types.py
```diff
@@ -43,25 +39,4 @@ class BaseLMBuddyConfig(
...
-            dir (str | Path, optional): Root path of the temporary directory
-
-        Returns:
-            Path to the temporary config file
-        """
-        with tempfile.TemporaryDirectory(dir=dir) as tmpdir:
-            config_path = Path(tmpdir) / name
-            self.to_yaml_file(config_path)
-            yield config_path
+    pass
```
veekaybee, line 42: do we need this if it's a pass?

veekaybee:
any README docs that need to change as a result of this? could you describe high-level the expected change in behavior, i.e. how does running CLI change, etc?

sfriedowitz:
>any README docs that need to change as a result of this? could you describe high-level the expected change in behavior, i.e. how does running CLI change, etc?

- Running the CLI does not change
- Ive updated CONTRIBUTING.md to account for the now-removed optional dependency groups

src/lm_buddy/cli/run.py
sfriedowitz: We could, but I somewhat like the syntax of calling `lm_buddy.run_job`. It feels natural as the sole function of the package.

src/lm_buddy/cli/run.py
sfriedowitz: Thanks! Although reminder that the internal imports were previously there for a purpose (to allow for executign portions of the library depending on which optional group was installed).

But the optional groups felt kind of silly after all given that the dependency set between eval/finetuning is roughhyl equivalent. So now that its removed, we have no reason to keep the internal imports.

src/lm_buddy/cli/schema.py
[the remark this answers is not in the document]
sfriedowitz: Ill end up adding a new example notebook showing how to execute the library directly, outside of Ray and the CLI.

src/lm_buddy/jobs/__init__.py
sfriedowitz: > we can consider dropping the args for clarity

What do you mean? I don't follow this.

src/lm_buddy/types.py
sfriedowitz: We don't technically. I can remove.

src/lm_buddy/jobs/__init__.py
veekaybee: if we don't expect anything other than `config` and `artifact_loader`, likely we don't need the `*`

sfriedowitz: Ahh, you mean making the artifact loader an arg instead of a kwarg. Sure, I can do that.

It's a bit of stylistic preference from old repos I've worked on, where we generally used kwargs as much as possible to make the call sites more clear. In this case, given that the `config` is the main argument and the `artifact_loader` is more of a dependency of the method, it felt natural.

But I don't mind much either way, ill change it.

pyproject.toml
```diff
@@ -17,6 +17,7 @@ license = "Apache-2.0"
 packages = [{ include = "lm_buddy", from = "src" }]
 
 [tool.poetry.dependencies]
+# Core
```
veekaybee, line 20: since this is a fairly large API change, bump the version of the package?

sfriedowitz: It's going into a dev branch, so a version bump will come at the end when merging into `main`.

veekaybee (approved):
LGTM to go into dev, let's add some more documentation, examples, and version management once it's ready to land

pyproject.toml
sfriedowitz: I've bumped it now actually, Ill leave the package as a 0.2.0 for the subsequent PRs

PR #55, opened 2024-02-16
RD2024-132: Reorganize layout of jobs directory

sfriedowitz:
Refactor of the `lm_buddy.jobs` module to place the configs and entrypoints collectively in their own folders. 

This makes the import structure a bit nicer for when we only want to import configs, and not the entrypoint functions.

No changes in functionality or CLI API.


src/lm_buddy/jobs/configs/base.py
```diff
@@ -0,0 +1,40 @@
+import contextlib
+import tempfile
+from pathlib import Path
+
+from pydantic_yaml import parse_yaml_file_as, to_yaml_file
+
+from lm_buddy.types import BaseLMBuddyConfig
+
```
binaryaaron, line 8: define an `__all__ = [...]` for your new files

binaryaaron (approved):
other than adding the `all`s, lgtm.

src/lm_buddy/jobs/configs/base.py
sfriedowitz: Ill add the __all__ to the configs folder.

PR #56, opened 2024-02-16
Consolidate workflow jobs into one job

binaryaaron:
As stated in the title. Simplifies the workflow and prevents a bunch of code duplication / starting a new container for each sub-test type.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -26,6 +26,7 @@ jobs:
           cmd: yq -oy '.tool.poetry.dev-dependencies.ruff' pyproject.toml
 
       - name: Install UV, venv, ruff
+        id: setup
         env:
           RUFF_VERSION: ${{ steps.get_ruff_version.outputs.result }}
         run: |
@@ -36,12 +37,14 @@ jobs:
         continue-on-error: false
 
       - name: Linting with Ruff
+        id: linting
         run: | 
           . .venv/bin/activate
           ruff --output-format=github .
         continue-on-error: false
 
       - name: Install Poetry and test dependencies
+        id: install_poetry_and_deps
         run: |
           . .venv/bin/activate
           uv pip install poetry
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -39,88 +39,28 @@ jobs:
...
-      - name: Install poetry
-        id: poetry
-        run: |
-          . .venv/bin/activate
-          uv pip install poetry 
         continue-on-error: false
 
-      - name: Lock Poetry
-        run: |
+      - name: Linting with Ruff
```
sfriedowitz, line 50: I think this step should be moved up above the `Poetry and dependencies step`.

Since we are installing Ruff individually its very fast and can give feedback on job status before the longer running Poetry install (or the unit/integration tests).

Should we also halt the entire workflow if Ruff fails, before kicking off the heavier testing steps?

.github/workflows/main.yaml
```diff
@@ -39,88 +39,28 @@ jobs:
...
+        run: | 
           . .venv/bin/activate
-          poetry lock
-
-      - name: Install test dependencies
+          ruff --output-format=github .
+        continue-on-error: true
+        
+      - name: Unit tests
+        id: unit_tests
```
sfriedowitz, line 57: Add an `id: integration_tests` below, or remove this line here?

.github/workflows/main.yaml
binaryaaron: sg! ty for the catch on the ordering.

.github/workflows/main.yaml
```diff
@@ -25,102 +25,42 @@ jobs:
...
-
-      - name: Install UV, venv
-        run: |
-          curl -LsSf https://astral.sh/uv/install.sh | sh
-          uv venv
-          echo "VIRTUAL_ENV=$HOME/.venv" >> $GITHUB_ENV
-
-      - name: Install Poetry
-        id: poetry
+      - name: Install Poetry and test dependencies
```
sfriedowitz, line 46:
```suggestion
      - name: Install Poetry and package
```

sfriedowitz (approved):
LGTM

PR #60, opened 2024-02-20
Fix publishing by sourcing venv differently

sfriedowitz:

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -16,12 +16,13 @@ jobs:
       runs-on: ubuntu-latest
       steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -67,7 +68,7 @@ jobs:
         # tag filter and label filter were already hit
         if: ${{ inputs.prod_or_test == 'prod' }}
         run: |
-            source "$HOME/venv/bin/activate"
+            . .venv/bin/activate
```
sfriedowitz, line 71: This is the actual fix, the rest just naming consistency

binaryaaron (approved):
:shipit:

PR #62, opened 2024-02-20
Update env var name for Poetry PyPI token

sfriedowitz:

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -18,10 +18,11 @@ permissions:
 
 jobs:
   build-docs:
+    name: Build Docs
     runs-on: ubuntu-latest
 
     steps:
-    - name: Checkout code
+    - name: Checkout Code
       uses: actions/checkout@v4
 
     - name: Set up Python 3.10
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yaml
```diff
@@ -61,9 +61,9 @@ jobs:
             poetry publish --repository testpypi
         continue-on-error: false
 
-      - name: Publish dist to real pypi
+      - name: Publish to PyPI
         env:
-            POETRY_PYPI_TOKEN: ${{ secrets.PYPI_KEY }}
+            POETRY_PYPI_TOKEN_PYPI: ${{ secrets.PYPI_KEY }}
```
sfriedowitz, line 66: This is the change in env var name according to the Poetry docs: https://python-poetry.org/docs/configuration/#pypi-tokenname

.github/workflows/docs.yaml
```diff
@@ -18,10 +18,11 @@ permissions:
 
 jobs:
   build-docs:
+    name: Build Docs
```
sfriedowitz, line 21: Adding names and consistent capitalization across steps for the different workflows

veekaybee (approved):

PR #64, opened 2024-02-20
RD2024-134: Setuptools build

sfriedowitz:
## What's changing

- Getting rid of Poetry as the build tool in favor of the standard setuptools builds
- Updating docs and actions to not use Poetry, using `uv` for pip installs in the actions

## How to test it

- Pull the branch and follow the contributing guide for editable installation via `pip`


.github/workflows/docs.yaml
```diff
@@ -25,21 +25,28 @@ jobs:
...
+        curl -LsSf https://astral.sh/uv/install.sh | sh
+        uv venv
+        echo "VIRTUAL_ENV=$HOME/.venv" >> $GITHUB_ENV
+      continue-on-error: false
+
+    - name: Install Docs Dependencies
+      id: install_build_dependencies
+      run: |
+        . .venv/bin/activate
+        uv pip install "lm-buddy[docs] @ ."
```
sfriedowitz, line 40: We pay a bit of a time cost here because `lm-buddy[docs]` also installs the base package. But it keeps the versions in sync with `pyproject.toml` as the source of truth.

Another approach could be to write out a `requirements-docs.txt` file in the repo and use dynamic metadata to pull that into `pyproject.toml`.

.github/workflows/docs.yaml
```diff
@@ -25,21 +25,28 @@ jobs:
     - name: Checkout Code
       uses: actions/checkout@v4
 
-    - name: Set up Python 3.10
```
sfriedowitz, line 28: Since we are using a `uv venv`, we don't need the Python toolchain setup in the GH actions.

.github/workflows/main.yaml
```diff
@@ -15,54 +15,46 @@ jobs:
...
-        env:
-          RUFF_VERSION: ${{ steps.get_ruff_version.outputs.result }}
         run: |
           curl -LsSf https://astral.sh/uv/install.sh | sh
           uv venv
-          uv pip install ruff
           echo "VIRTUAL_ENV=$HOME/.venv" >> $GITHUB_ENV
         continue-on-error: false
 
+      - name: Install Package
```
sfriedowitz, line 26: Since the ruff version parsing was not working, I'm just installing the full thing before linting. It's quicker with `uv` at least 🤷

.github/workflows/publish.yaml
```diff
@@ -53,23 +48,25 @@ jobs:
           retention-days: 1
 
       - name: Publish to Test PyPI
-        if: ${{ inputs.prod_or_test == 'test' }}
+        if: ${{ inputs.prod_or_test == 'Test' }}
         env:
-            POETRY_PYPI_TOKEN_TESTPYPI: ${{ secrets.PYPI_TEST_KEY }}
+            TWINE_USERNAME: "__token__"
+            TWINE_PASSWORD: ${{ secrets.PYPI_TEST_KEY }}
```
sfriedowitz, line 54: Not positive this works, but probably can't test it until we merge

.github/workflows/docs.yaml
binaryaaron: oh nice; need to make sure the venv is created with python 3.10

.github/workflows/docs.yaml
```diff
@@ -25,21 +25,28 @@ jobs:
...
-      uses: actions/setup-python@v5
-      with:
-        python-version: "3.10"
+    - name: Set up Environment
+      id: setup
+      run: |
+        curl -LsSf https://astral.sh/uv/install.sh | sh
+        uv venv
```
binaryaaron, line 32: double check the install here is for version 3.10 if removing above

.github/workflows/docs.yaml
binaryaaron: this is related to the pypa/pip issue you shared in slack?

.github/workflows/main.yaml
```diff
@@ -15,54 +15,46 @@ jobs:
...
-          cmd: yq -oy '.tool.poetry.dev-dependencies.ruff' pyproject.toml
-
-      - name: Install uv, venv, Ruff
+      - name: Set up Environment
         id: setup
-        env:
-          RUFF_VERSION: ${{ steps.get_ruff_version.outputs.result }}
         run: |
           curl -LsSf https://astral.sh/uv/install.sh | sh
           uv venv
```
binaryaaron, line 22: same as above, make sure the py version is right

.github/workflows/main.yaml
binaryaaron: this is due to changing the format from `ruff = "0.2.1"` to the ` ruff = [ruff...]` i take it? YQ should be able to parse that, can test in a terminal.

binaryaaron: image
works locally with this changeset. what was wrong?

.github/workflows/docs.yaml
sfriedowitz: It is, yeah. There is no way I can find to install only the extras. 

Same goes for Ruff, and how you tried to parse the toml section to get around that.

Using pyproject alone without requirements files is lacking here a bit.

.github/workflows/docs.yaml
sfriedowitz: Hmm, yeah, maybe I should add it back. it only can create a venv from a Python interpreter on the system path. So I should ensure its there on the GH actions runner.

Think I got lucky in that it worked while removing it, their base env must just have 3.10

.github/workflows/main.yaml
sfriedowitz: I updated the `yq` command to work with the new pyproject format. 

The `steps.get_ruff_version.outputs.result` value is `null` between steps, and I don't know why.

The `RUFF_VERSION` env var was not being used before my changes, which is why we missed that.

sfriedowitz: Hmm, the `null` appears to be the output from `yq` for a no-match.

I wonder if we are missing the match for some reason when running in GH actions.

.github/workflows/main.yaml
```diff
@@ -20,49 +20,45 @@ jobs:
...
+        id: get_ruff_version
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq -oy '.project.optional-dependencies.ruff[0]' pyproject.toml
+
       - name: Linting with Ruff
         id: linting
         run: | 
           . .venv/bin/activate
+          uv pip install ${{ steps.get_ruff_version.outputs.result }}
```
sfriedowitz, line 41: It's working now 🤷

binaryaaron (approved):
approved; @sfriedowitz feel free to merge if you like the latest changes i added. 

basically got rid of YQ in favor of using a cli toml parser via python; simplifies it and runs the setup+linting in about 1s.

sfriedowitz:
> basically got rid of YQ in favor of using a cli toml parser via python; simplifies it and runs the setup+linting in about 1s.

Confirmed is working, thanks!

PR #70, opened 2024-02-22
Fix up epoch vs epochs in config

chainlink:
Noticed `epochs` was an invalid value when running training.


sfriedowitz:
Good catch, add a patch version bump and this is g2g!

sfriedowitz (approved):
LGTM

PR #71, opened 2024-02-22
Using Fine Tuning using Ray submission CLI

chainlink:
This adds some (very light) docs on using the ray job submission CLI to submit a finetuning job


sfriedowitz:
This is essentially already stated in the package Readme. I think this is duplicate?

chainlink:
The difference here is that there is no python driver script required, it's all CLI based. I can add to the Readme though.

PR #73, opened 2024-02-29
PR check to enforce version bump

sfriedowitz:
## What's changing

Adds a PR check that enforces a version bump on the package, by comparing to the version in the current `main` branch. This will enforce developers to bump before merging any code to `main`.


.github/workflows/pr_checks.yaml
```diff
@@ -0,0 +1,120 @@
...
+        continue-on-error: false
+
+      - name: Unit Tests
+        id: unit_tests
+        run: |
+          . .venv/bin/activate
+          pytest tests/unit
+        continue-on-error: false
+
+      - name: Integration Tests
```
veekaybee, line 45: would this mean that we run integration tests twice times when we merge into main as this is activated since we would now have two workflows that run against main?

veekaybee: oh, just saw we deleted the main workflow here and are replacing, ignore

.github/workflows/pr_checks.yaml
```diff
@@ -0,0 +1,120 @@
...
+        continue-on-error: false
+
+      - name: Install Dependencies
+        id: install_dependencies
+        run: |
+          . .venv/bin/activate
+          uv pip install toml-cli packaging
+        continue-on-error: false
+
+      - name: Validate Version Bump
```
veekaybee, line 112: could we use something like this instead since it already exists or are there features that are missing? https://github.com/marketplace/actions/require-semantic-versioning-bump

sfriedowitz: Good question, I didn't know that exists. Let me check later if it will work.

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "lm-buddy"
-version = "0.2.5"
+version = "0.2.6"
```
binaryaaron, line 7: related - do projects with the toml project file also have a `version.py` file somewhere?

sfriedowitz: Not necessarily. It is possible to define a `version.py` and dynamically load that into the `pyproject.toml`, but I have chosen to avoid that for now.

.github/workflows/version.yaml
```diff
@@ -0,0 +1,19 @@
+name: Python Version
+
+on:
+  pull_request:
+    branches: [main]
```
sfriedowitz, line 5: Turned this into its own workflow so it can only be triggered on PRs into `main`, rather than dev branches

sfriedowitz:
Once this is approved, I will change the branch protection rules to check for the new GH actions instead of the old name.

.github/workflows/pr_checks.yaml
```diff
@@ -1,4 +1,4 @@
-name: Tests
+name: PR Checks
```
veekaybee, line 1: Maybe something more specific, aka main merge checks or similar?

veekaybee (approved):
LGTM with a few small comments

PR #75, opened 2024-02-29
Added v0 of prometheus lm-buddy entrypoint

aittalam:
Here's the first version of the lm-buddy prometheus entrypoint!

I tried to keep the script relatively close to the original evaluation script, as I'd like to use it as a use case for a "conversion" of some existing tool (a tutorial is being written...)

I feel it still needs some massaging as I am not 100% sure of the implementation choices I took (format of configs, using one method or another to get artifacts, etc). Happy to receive some early feedback and iterate on it soon!

### DIFF
--- a/examples/configs/prometheus/prometheus_config.yaml
+++ b/examples/configs/prometheus/prometheus_config.yaml
@@ -4,6 +4,7 @@ dataset:
     version: "latest"
     project: "lm-buddy-prometheus"
     entity: "mozilla-ai"
+  # field containing scoring instructions in the json file
   text_field: "instruction"
 
 prometheus:
@@ -12,6 +13,7 @@ prometheus:
   tokenizer:
     load_from: "meta-llama/Llama-2-7b-chat-hf"
   max_tokens: 256
+  # number of times the model is called per sample
   num_answers: 3
 
 tracking:
--- a/examples/configs/prometheus/prometheus_config.yaml
+++ b/examples/configs/prometheus/prometheus_config.yaml
@@ -1,0 +1,20 @@
+dataset:
+  load_from:
+    name: "wandb_file_artifact_name.json"
+    version: "latest"
+    project: "lm-buddy-prometheus"
+    entity: "mozilla-ai"
+  text_field: "instruction"
+
+prometheus:
+  inference:
+    base_url: "http://your.vllm.server:8000/v1"
+  tokenizer:
+    load_from: "meta-llama/Llama-2-7b-chat-hf"
+  max_tokens: 256
+  num_answers: 3
+
+tracking:
+  name: "lm-buddy-prometheus"
+  project: "lm-buddy-examples"
+  entity: "mozilla-ai"
--- a/examples/configs/prometheus/prometheus_config.yaml
+++ b/examples/configs/prometheus/prometheus_config.yaml
@@ -10,11 +10,24 @@ dataset:
 prometheus:
   inference:
     base_url: "http://your.vllm.server:8000/v1"
-  tokenizer:
-    load_from: "meta-llama/Llama-2-7b-chat-hf"
-  max_tokens: 256
-  # number of times the model is called per sample
+    engine: "kaist-ai/prometheus-13b-v1.0"
+  best_of: 1
+  max_tokens: 512
+  frequency_penalty: 1.03
+  temperature: 1.0
+  top_p: 0.9
+
+evaluation:
+  # number of times a model is evaluated per sample
   num_answers: 3
+  # max number of retries if a communication error
+  # with the server occurs
+  max_retries: 5
+  # min and max scores as defined in the scoring rubric
+  min_score: 1
+  max_score: 5
+  # enable/disable tqdm to track eval progress
+  enable_tqdm: True
 
 tracking:
   name: "lm-buddy-prometheus"
--- a/src/lm_buddy/cli/run.py
+++ b/src/lm_buddy/cli/run.py
@@ -1,7 +1,12 @@
 import click
 
 import lm_buddy
-from lm_buddy.jobs.configs import FinetuningJobConfig, LMHarnessJobConfig, SimpleJobConfig
+from lm_buddy.jobs.configs import (
+    FinetuningJobConfig,
+    LMHarnessJobConfig,
+    PrometheusJobConfig,
+    SimpleJobConfig,
+)
 
 # TODO(RD2024-125): We should probably collapse all these commands into a single CLI command
 # - Need to figure out best way to polymorphically deserialize the job config classes
@@ -31,4 +36,11 @@ def run_finetuning(config: str) -> None:
 @click.option("--config", type=str)
 def run_lm_harness(config: str) -> None:
     config = LMHarnessJobConfig.from_yaml_file(config)
+    lm_buddy.run_job(config)
+
+
+@group.command("prometheus", help="Run the prometheus evaluation job.")
+@click.option("--config", type=str)
+def run_prometheus(config: str) -> None:
+    config = PrometheusJobConfig.from_yaml_file(config)
     lm_buddy.run_job(config)
--- a/src/lm_buddy/integrations/vllm.py
+++ b/src/lm_buddy/integrations/vllm.py
@@ -12,8 +12,26 @@ class InferenceServerConfig(BaseLMBuddyConfig):
 
     Note: This configuration is intended to be generic and not bound to the interface
     of any specific training/evaluation framework. See `LocalChatCompletionConfig`
-    for intended usage alongside a third-party framework.
+    or `vLLMCompleptionsConfig` for intended usage alongside a third-party framework.
     """
 
     base_url: str
     engine: str | HuggingFaceAssetPath | None = None
+
+
+class vLLMCompletionsConfig(BaseLMBuddyConfig):
+    """Configuration for a vLLM-based completions service
+
+    The "local-chat-completions" model is powered by a self-hosted inference server,
+    specified as an `InferenceServerConfig`. Additional arguments are also provided
+    to control the tokenizer type and generation parameters.
+    """
+
+    inference: InferenceServerConfig
+
+    # vLLM-specific params
+    best_of: int | None = None
+    max_tokens: int | None = None
+    frequency_penalty: float | None = None
+    temperature: float | None = None
+    top_p: float | None = None
--- a/src/lm_buddy/integrations/vllm.py
+++ b/src/lm_buddy/integrations/vllm.py
@@ -19,7 +19,7 @@ class InferenceServerConfig(BaseLMBuddyConfig):
     engine: str | HuggingFaceAssetPath | None = None
 
 
-class vLLMCompletionsConfig(BaseLMBuddyConfig):
+class VLLMCompletionsConfig(BaseLMBuddyConfig):
     """Configuration for a vLLM-based completions service
 
     The "local-chat-completions" model is powered by a self-hosted inference server,
--- a/src/lm_buddy/integrations/vllm.py
+++ b/src/lm_buddy/integrations/vllm.py
@@ -25,6 +25,11 @@ class VLLMCompletionsConfig(BaseLMBuddyConfig):
     The "local-chat-completions" model is powered by a self-hosted inference server,
     specified as an `InferenceServerConfig`. Additional arguments are also provided
     to control the tokenizer type and generation parameters.
+
+    Note that this is just a subset of the parameters allowed by a vLLM server (see
+    https://github.com/vllm-project/vllm/blob/main/vllm/sampling_params.py). If we
+    choose to use this configuration to cover for more use cases, it will make sense
+    to add the other supported configuration parameters too.
     """
 
     inference: InferenceServerConfig
--- a/src/lm_buddy/integrations/wandb/artifact_utils.py
+++ b/src/lm_buddy/integrations/wandb/artifact_utils.py
@@ -4,7 +4,7 @@ from typing import Any
 from urllib.parse import ParseResult, urlparse
 
 import wandb
-import os
+
 
 class ArtifactType(str, Enum):
     """Enumeration of artifact types used by the LM Buddy."""
--- a/src/lm_buddy/integrations/wandb/artifact_utils.py
+++ b/src/lm_buddy/integrations/wandb/artifact_utils.py
@@ -4,7 +4,7 @@ from typing import Any
 from urllib.parse import ParseResult, urlparse
 
 import wandb
-
+import os
 
 class ArtifactType(str, Enum):
     """Enumeration of artifact types used by the LM Buddy."""
@@ -109,4 +109,39 @@ def build_table_artifact(
     for table_name, table_data in tables.items():
         table = wandb.Table(data=table_data, columns=columns)
         artifact.add(table, name=table_name)
+    return artifact
+
+
+def build_file_artifact(
+    artifact_name: str,
+    artifact_type: ArtifactType,
+    file_path: str | Path,
+    *,
+    reference: bool = False,
+    entry_name: str | None = None,
+) -> wandb.Artifact:
+    """Build an artifact containing a single file
+
+    Args:
+        artifact_name (str): Name of the artifact
+        artifact_type (ArtifactType): Type of artifact
+        file_path (str | Path): The full path (including filename) of the file
+
+    Keyword Args:
+        reference (bool): Only reference the file, do not copy contents. Defaults to False.
+        entry_name (str | None): Name for the file within the artifact. If None, defaults
+                                 to the original filename.
+
+    Returns:
+        wandb.Artifact: The generated artifact.
+    """
+    artifact = wandb.Artifact(name=artifact_name, type=artifact_type)
+
+    if reference:
+        artifact.add_reference(
+            uri=f"{ArtifactURIScheme.FILE}://{file_path}",
+            name=entry_name,
+        )
+    else:
+        artifact.add_file(str(file_path), name=entry_name)
     return artifact
--- a/src/lm_buddy/jobs/__init__.py
+++ b/src/lm_buddy/jobs/__init__.py
@@ -1,9 +1,15 @@
 from lm_buddy.integrations.wandb import ArtifactLoader, WandbArtifactLoader
-from lm_buddy.jobs._entrypoints import run_finetuning, run_lm_harness, run_simple
+from lm_buddy.jobs._entrypoints import (
+    run_finetuning,
+    run_lm_harness,
+    run_prometheus,
+    run_simple,
+)
 from lm_buddy.jobs.configs import (
     FinetuningJobConfig,
     LMBuddyJobConfig,
     LMHarnessJobConfig,
+    PrometheusJobConfig,
     SimpleJobConfig,
 )
 
@@ -26,5 +32,7 @@ def run_job(
             run_finetuning(finetuning_config, artifact_loader)
         case LMHarnessJobConfig() as lm_harness_config:
             run_lm_harness(lm_harness_config, artifact_loader)
+        case PrometheusJobConfig() as prometheus_config:
+            run_prometheus(prometheus_config, artifact_loader)
         case _:
             raise ValueError(f"Received invalid job configuration: {config}")
--- a/src/lm_buddy/jobs/_entrypoints/prometheus.py
+++ b/src/lm_buddy/jobs/_entrypoints/prometheus.py
@@ -14,7 +14,7 @@ from transformers import AutoTokenizer
 from openai import OpenAIError, OpenAI
 
 from tqdm import tqdm
-import os
+from pathlib import Path
 import json
 import copy
 
@@ -67,7 +67,7 @@ def run_prometheus(config: PrometheusJobConfig, artifact_loader: ArtifactLoader)
     # load dataset from W&B artifact
     hf_loader = HuggingFaceAssetLoader(artifact_loader)
     artifact_path,_ = hf_loader.resolve_asset_path(config.dataset.load_from)
-    dataset_fname = os.path.join(artifact_path, config.dataset.load_from.name)
+    dataset_fname = Path(artifact_path) / config.dataset.load_from.name
     
     with open(dataset_fname,'r') as f:
         # eval samples are JSON-encoded, each takes one line in the dataset file
@@ -82,9 +82,9 @@ def run_prometheus(config: PrometheusJobConfig, artifact_loader: ArtifactLoader)
     )
 
     # open the output file for writing and iterate on samples
-    output_fname = os.path.join("/tmp", config.tracking.name)
+    output_fname = Path("/tmp") / config.tracking.name
     with open(output_fname,'w') as file:
-        for sample in tqdm(data):
+        for sample in tqdm(data[:1]):
             # convert instructions from the dataset (`text_field` in a dict) to
             # prompts that prometheus accepts
             prompt = instruction_to_prompt(sample[config.dataset.text_field])
--- a/src/lm_buddy/jobs/_entrypoints/prometheus.py
+++ b/src/lm_buddy/jobs/_entrypoints/prometheus.py
@@ -12,7 +12,7 @@ from lm_buddy.integrations.wandb import (
 )
 from fastchat.conversation import get_conv_template
 from transformers import AutoTokenizer
-from openai import OpenAIError, OpenAI
+from openai import OpenAIError, OpenAI, Completion
 
 from tqdm import tqdm
 from pathlib import Path
@@ -25,7 +25,15 @@ class BadResponseException(Exception):
         self.error = error
 
 
-def openai_completion(config, client, prompt):
+def openai_completion(
+    config: PrometheusJobConfig, 
+    client: OpenAI, 
+    prompt: str
+) -> Completion:
+    """ Connects to a remote OpenAI-API-compatible Prometheus endpoint
+        and returns a Completion holding the model's response.
+    """
+
     return client.completions.create(
         model = config.prometheus.inference.engine,
         prompt = prompt,
@@ -37,7 +45,15 @@ def openai_completion(config, client, prompt):
     )
 
 
-def parse_response(config, response):
+def parse_response(
+    config: PrometheusJobConfig, 
+    response: Completion
+) -> tuple[str, str]:
+    """ Given a Prometheus eval response as returned by the OpenAI API
+        endpoint (i.e. in Completion format), extract feedback
+        and score.
+    """
+    
     if response is None:
         raise BadResponseException("Server returned an empty response")
 
@@ -58,7 +74,12 @@ def parse_response(config, response):
     return feedback, score
 
 
-def instruction_to_prompt(instruction):
+def instruction_to_prompt(
+    instruction: str
+) -> str:
+    """ Given some text containing Prometheus instructions, transform it
+        into an actual llama-2 prompt.
+    """
     conv = get_conv_template("llama-2")
     conv.set_system_message("You are a fair evaluator language model.")
     conv.append_message(conv.roles[0], instruction)
--- a/src/lm_buddy/jobs/_entrypoints/prometheus.py
+++ b/src/lm_buddy/jobs/_entrypoints/prometheus.py
@@ -1,84 +1,75 @@
 # lm-buddy entrypoint to run evaluations using a Prometheus inference server
 # see https://github.com/kaistAI/prometheus/blob/main/evaluation/benchmark/run_absolute_scoring.py
 
-from lm_buddy.jobs.configs import PrometheusJobConfig
+import copy
+import json
+from pathlib import Path
+
+from fastchat.conversation import get_conv_template
+from openai import Completion, OpenAI, OpenAIError
+from tqdm import tqdm
+
 from lm_buddy.integrations.huggingface import HuggingFaceAssetLoader
 from lm_buddy.integrations.huggingface.tokenizer_config import AutoTokenizerConfig
 from lm_buddy.integrations.wandb import (
-    ArtifactType, 
-    ArtifactLoader, 
-    build_file_artifact, 
+    ArtifactLoader,
+    ArtifactType,
+    build_file_artifact,
     wandb_init_from_config,
 )
-from fastchat.conversation import get_conv_template
-from transformers import AutoTokenizer
-from openai import OpenAIError, OpenAI, Completion
+from lm_buddy.jobs.configs import PrometheusJobConfig
 
-from tqdm import tqdm
-from pathlib import Path
-import json
-import copy
 
-class BadResponseException(Exception):
+class BadResponseError(Exception):
     def __init__(self, message, error=None):
         self.message = message
         self.error = error
 
 
-def openai_completion(
-    config: PrometheusJobConfig, 
-    client: OpenAI, 
-    prompt: str
-) -> Completion:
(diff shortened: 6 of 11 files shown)


src/lm_buddy/integrations/wandb/artifact_utils.py
```diff
@@ -110,3 +110,38 @@ def build_table_artifact(
         table = wandb.Table(data=table_data, columns=columns)
         artifact.add(table, name=table_name)
     return artifact
+
+
+def build_file_artifact(
```
aittalam, line 115: I kept a structure very similar to `build_directory_artifact` and this might seem a bit redundant, so I am happy to think about alternatives.

src/lm_buddy/jobs/configs/prometheus.py
```diff
@@ -0,0 +1,42 @@
...
+    inference: InferenceServerConfig
+
+    # vLLM-served model params
+    best_of: int = 1
+    max_tokens: int = 512 
+    frequency_penalty: float = 1.03
+    temperature: float = 1.0
+    top_p: float = 0.9
+
+    # evaluation script params
```
aittalam, line 28: This section could become its own config, while the previous one could be general for all vLLM-served models. WDYT?

sfriedowitz: I think it depends. Which of these parameters are general for _any_ calls to a vLLM inference server, and which are specific to the Prometheus code? 

It seems like some of these are the general HuggingFace inference parameters (top_k, repetition_penalty, etc..). Perhaps those should exist on their own config like `GenerationConfig` or something?

sfriedowitz (commented):
A handful of comments, some bigger than others. Happy to discuss more or chat offline if anything didn't make sense.

src/lm_buddy/jobs/__init__.py
```diff
@@ -1,9 +1,15 @@
 from lm_buddy.integrations.wandb import ArtifactLoader, WandbArtifactLoader
-from lm_buddy.jobs._entrypoints import run_finetuning, run_lm_harness, run_simple
+from lm_buddy.jobs._entrypoints import (
+    run_finetuning,
+    run_lm_harness,
```
veekaybee, line 4: the same question here as above, we now have finetuning and eval that are on the same level, and harness and prometheus that are a subset of eval, do we want to handle in a nested hierarchy?

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+from fastchat.conversation import get_conv_template
+from transformers import AutoTokenizer
+from openai import OpenAIError, OpenAI
+
+from tqdm import tqdm
+import os
+import json
+import copy
+
+class BadResponseException(Exception):
```
veekaybee, line 21: what does the bad response mean in this case, is it an error from the server? we might want to specify the exception type a bit more here

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+    def __init__(self, message, error):
+        self.message = message
+        self.error = error
+
+
+def openai_completion(config, client, prompt):
```
veekaybee, line 27: let's add type hints/return types (and the same for all the other new methods)

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+    return client.completions.create(
+        model = "kaist-ai/prometheus-13b-v1.0",
```
veekaybee, line 29: we should pass this in through config I think

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+    # get the tokenizer
+    tokenizer = hf_loader.load_pretrained_tokenizer(config.prometheus.tokenizer)
+
+    # instantiate OpenAI client to speak with the vLLM endpoint
+    client = OpenAI(
+        base_url = config.prometheus.inference.base_url
+    )
+
+    # open the output file for writing and iterate on samples
+    output_fname = os.path.join("/tmp", config.tracking.name)
```
veekaybee, line 81: let's consider using Pathlib since we mostly use that across the library to standardize

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+    with open(output_fname,'w') as file:
+        for sample in tqdm(data):
```
veekaybee, line 83: does tqdm work the same way in ray logs as regularly? might be something to check here if we're planning on displaying this output in the Ray Dashboard logs

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+            result['prometheus_score'] = []
+
+            for idx in range(config.prometheus.num_answers):
+
+                i = 0
+                while i < config.prometheus.max_retries: 
+                    try:
+                        response = openai_completion(config, client, prompt)
+                        feedback, score = parse_response(response)
+                        print(feedback, score)
```
veekaybee, line 105: do we need to print these?

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+        best_of = config.prometheus.best_of,
+        max_tokens = config.prometheus.max_tokens,
+        frequency_penalty = config.prometheus.frequency_penalty,
+        temperature = config.prometheus.temperature,
+        top_p = config.prometheus.top_p
+    )
+
+
+def parse_response(response):
+    try:
```
veekaybee, line 40: what do you think about changing this to an if statement since we generally reserve assert for tests? so

```
if response and len(response) > 0:
   response_text.....
```

src/lm_buddy/jobs/configs/prometheus.py
aittalam: Exactly, that was my thought. All of the params in that section are vLLM params (which, despite serving the same HF models, might differ from other inference servers depending on the API they implement).
So my thought was that we could have these (and potentially other) parameters for a vLLMGenerationConfig and import it in the Prometheus-specific config. The only thing we miss is the defaults (I liked the idea of having the values used by kaistai's team as defaults here), but I think it's ok - we can just provide a ready-to-use configuration file with those defaults in.

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+    with open(output_fname,'w') as file:
+        for sample in tqdm(data):
+            # convert instructions from the dataset (`text_field` in a dict) to
+            # prompts that prometheus accepts
+            prompt = instruction_to_prompt(sample[config.dataset.text_field])
+
+            # skip those examples which are too long 
+            tokenized_prompt  = tokenizer(prompt, truncation=False)
+            if(len(tokenized_prompt['input_ids'])>3072):
+                continue
```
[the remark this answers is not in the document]
aittalam: Good point, this was laziness on my side trying to replicate kaistai's code as much as possible and yes, that's the only place where it is used.

I think as long as the inference server (possibly not just vLLM - I am thinking e.g. llamafile) is able to, we could use its tokenizer. We can have the tokenizer parameter as optional and if it is not provided we'll use the service to tokenize, otherwise do it locally. Wdyt?

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+    conv.append_message(conv.roles[1], None)
+    return conv.get_prompt()
+
+
+def run_prometheus(config: PrometheusJobConfig, artifact_loader: ArtifactLoader):
+
+    # load dataset from W&B artifact
+    hf_loader = HuggingFaceAssetLoader(artifact_loader)
+    artifact_path,_ = hf_loader.resolve_asset_path(config.dataset.load_from)
+    dataset_fname = os.path.join(artifact_path, config.dataset.load_from.name)
```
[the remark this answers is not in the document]
aittalam: Oh yes, I kinda misused the class because I did not want to create another one specifically for JSON files 😅  (bad, I know, but I think we should expect this kind of behavior from lm-buddy users and either embrace the "there's more than one way to do it" paradigm or provide more specific classes, e.g. HFTextDatasetConfig for HF datasets and TextDatasetConfig for other file types).
For the dataset specifically, I agree. I like that it works on those original JSON files though. I'd like to make a new entrypoint specifically for preprocessing that will take the JSON files and save them as HF dataset artifacts (perhaps in next PR). This can also be a good example to add to our documentation too!

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,131 @@
...
+                result['prometheus_score'].append(score)
+
+            # dump sample results
+            file.write(json.dumps(result)+"\n")
+
+
+    # Register a dataset file artifact if tracking is enabled
+    if config.tracking:
+        
+        with wandb_init_from_config(config.tracking) as run:
```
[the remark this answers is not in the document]
aittalam: Great point, I'll do that!

src/lm_buddy/cli/run.py
```diff
@@ -32,3 +37,10 @@ def run_finetuning(config: str) -> None:
 def run_lm_harness(config: str) -> None:
     config = LMHarnessJobConfig.from_yaml_file(config)
     lm_buddy.run_job(config)
+
+
+@group.command("prometheus", help="Run the prometheus evaluation job.")
```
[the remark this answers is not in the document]
aittalam: Good point! And wdyt if instead of just changing the filenames we change the CLI groups? 
E.g. `lm_buddy run eval lm-harness`, `lm_buddy run eval ragas`, etc.
Actually `lm_buddy finetune ...`, `lm_buddy eval ...` sound even better, but I am not sure whether this would break anything

src/lm_buddy/jobs/_entrypoints/prometheus.py
aittalam: So, this is more like a Prometheus quirk. The server can return errors (`OpenAIError`) which are trapped out of this class. But sometimes, even if the server response arrives it might be malformed - e.g. when `max_tokens` is too short for Prometheus to return a full answer (so you end up without a score, hence the split error) or when the values for the score are not valid... Which reminds me this has to be a parameter if we want to customize our rubrics! Good catch, thanks for raising this :-)

src/lm_buddy/jobs/_entrypoints/prometheus.py
aittalam: Yes! Right now it is hardcoded as we do not have many alternatives, but we can expect future Prometheus-compatible models

src/lm_buddy/jobs/_entrypoints/prometheus.py
aittalam: Yesterday was all local testing, ray is next!
My intuition is it will be a bit ugly (as it usually happens with other tqdm-based iterations such as loading models or trainings) but should not break stuff.
I'll also make sure the logging is a bit less verbose than now (e.g. I could skip printing all the answers, which are on rn for debugging in terminal)

src/lm_buddy/jobs/_entrypoints/prometheus.py
aittalam: ahahah no 😅 , will fix it!

aittalam:
Many thanks for your comments! I will go through them (marked all of the actionable ones with a 👍) and also fix the linting issues before asking for approval 🙏

src/lm_buddy/cli/run.py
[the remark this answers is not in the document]
sfriedowitz: I would say if we are going to change it, we should just go with "lm-buddy finetune" and "lm-buddy evaluate ragas/prometheus/lm-harness".

But then how does the "simple" job fit in semantically? We can probably just delete that since I was just using it for initial debugging. 

No need to cram these changes into this PR, though. We can make a ticket for it.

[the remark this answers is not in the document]
veekaybee: Agree, I think it would be good to put the simple job in a "cookbook" if we decide to separate the examples or in the documentation and take it out of the codebase

src/lm_buddy/jobs/configs/prometheus.py
sfriedowitz: > The only thing we miss is the defaults

I don't follow, why would we be missing the defaults if we created a `vLLMGenerationConfig` class? We could set the defaults there.

Anyhow, I think having those parameters bundled into a re-usable Pydantic model makes sense.

src/lm_buddy/jobs/_entrypoints/prometheus.py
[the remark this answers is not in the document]
sfriedowitz: I think in general, any externally hosted model will be bundled with its appropriate tokenizer (am I wrong here??). So when you interact with that model's API, you just send text and it handles the tokenization internally.

If we can rely on that here, I would suggest we remove the tokenizer option all together on these configs. It seems like it would be possible for a user to specify a different tokenizer than what is running on the vLLM server, leading to unexpected/erroneous results, so I would like to avoid that if possible.

src/lm_buddy/jobs/_entrypoints/prometheus.py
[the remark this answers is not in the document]
sfriedowitz: I think for now we should rely on the existing asset load mechanism for HF datasets, since the configs and associated methods were geared towards that. I think it would be simplify this code alot. 

But I agree in general that supporting other dataset types could be a long term goal, and may require a more holistic refactor of the dataset loading functions.

>I'd like to make a new entrypoint specifically for preprocessing that will take the JSON files and save them as HF dataset artifacts (perhaps in next PR).

Sure! We can add some helper functions here, and perhaps add a section on that to the `dataset_preprocessing` example notebook that already exists. The HuggingFace `Dataset` class contains some extra constructors methods for instantiating their class from various data formats, I would check that out.

src/lm_buddy/jobs/_entrypoints/prometheus.py
sfriedowitz: Ruff will enforce that we do this :D

src/lm_buddy/jobs/_entrypoints/prometheus.py
sfriedowitz: FWIW, in the finetuning code I set `disable_tqdm=True` in the training args because I think it shows up malformatted in the running logs.

src/lm_buddy/jobs/configs/prometheus.py
aittalam: What I meant is if we have a `vLLMGenerationConfig` and we assume to use it generally for vLLM models, then it is better not to define it with Prometheus' specific defaults. But that's ok, I'll go with the reusable model and an explicit config

examples/configs/prometheus/prometheus_config.yaml
```diff
@@ -0,0 +1,22 @@
...
+    name: "wandb_file_artifact_name.json"
+    version: "latest"
+    project: "lm-buddy-prometheus"
+    entity: "mozilla-ai"
+  # field containing scoring instructions in the json file
+  text_field: "instruction"
+
+prometheus:
+  inference:
+    base_url: "http://your.vllm.server:8000/v1"
```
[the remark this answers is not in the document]
aittalam: That's a good question. Right now I am replicating what I saw in the other entrypoints, but we can consider using env variables in a later PR (I'd do that as an additional alternative, not in substitution of the current approach, though, so users can choose what works best for them)

src/lm_buddy/jobs/_entrypoints/prometheus.py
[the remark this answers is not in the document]
aittalam: Following Vicki's comment, I now allow for an `engine` parameter which is at the same time the model used on the vLLM server and the local tokenizer used to calculate prompt length before sending to the model. This way there is no discrepancy between the two models.

src/lm_buddy/jobs/_entrypoints/prometheus.py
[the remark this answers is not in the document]
aittalam: Added load/save of datasets in HF format!
The output is a file that is updated incrementally (you can just `tail` it to follow its progress and if the process breaks you have not lost everything), and at the end it is synced with wandb if tracking is enabled.
The code to load the json file and save it as HF dataset is literally two lines but I am happy to add it to documentation in a follow-up!

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
...
+    ArtifactLoader,
+    ArtifactType,
+    build_directory_artifact,
+    wandb_init_from_config,
+)
+from lm_buddy.jobs.common import LMBuddyJobType
+from lm_buddy.jobs.configs import PrometheusJobConfig
+
+
+class BadResponseError(Exception):
```
veekaybee, line 26: since this class doesn't have any methods but only handles variables, it could be a dataclass instead:

```python
from dataclasses import dataclass

@dataclass
class BadResponseError(Exception):
    message: str
    error: str = None
```

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
...
+    if response is None:
+        raise BadResponseError("Server returned an empty response")
+
+    try:
+        response_text = response.choices[0].text
+        # note: this can raise a ValueError if the message is malformed
+        feedback, score = response_text.split("[RESULT]")
+        feedback = feedback.strip()
+        score = score.strip()
+        if score not in [
```
veekaybee, line 63: if this score will always be a string list of [0,1,2,3,4,5], it might be more legible to specify it as such in the config: 

```python
config.evaluation.scores = [0,1,2,3,4,5]

if int(score) not in config.evaluation.scores:
     raise BadResponseError(f"Score {score} is not in range")
```

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
...
+    output_fname = Path(config.evaluation.tmp_folder) / tracking_name
+    with output_fname.open("w") as file:
+        for sample in dataset_iterable:
+            # convert instructions from the dataset (`text_field` in a dict) to
+            # prompts that prometheus accepts
+            prompt = instruction_to_prompt(sample[config.dataset.text_field])
+
+            # skip those examples which are too long
+            tokenized_prompt = tokenizer(prompt, truncation=False)
+            if len(tokenized_prompt["input_ids"]) > 3072:
```
veekaybee, line 101: should `3072` also be set in the prometheus config?

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
...
+                continue
+
+            # prepare output
+            result = copy.deepcopy(sample)
+            result["prometheus_output"] = []
+            result["prometheus_score"] = []
+
+            for idx in range(config.evaluation.num_answers):
+                i = 0
```
veekaybee, line 110: could we rename `i` to something more descriptive, like `current_retry_attempt` for troubleshooting?

src/lm_buddy/integrations/vllm.py
```diff
@@ -12,8 +12,26 @@ class InferenceServerConfig(BaseLMBuddyConfig):
...
+    """Configuration for a vLLM-based completions service
+
+    The "local-chat-completions" model is powered by a self-hosted inference server,
+    specified as an `InferenceServerConfig`. Additional arguments are also provided
+    to control the tokenizer type and generation parameters.
+    """
+
+    inference: InferenceServerConfig
+
+    # vLLM-specific params
```
veekaybee, line 32: do these come from here? we might want to add a note in case we decide to add more later https://github.com/vllm-project/vllm/blob/main/vllm/sampling_params.py

examples/configs/prometheus/prometheus_config.yaml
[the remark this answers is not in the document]
sfriedowitz: Everything is specified by configs right now, and we aren't leaning on env vars in the package. I would suggest we maintain this pattern for the time being.

src/lm_buddy/cli/run.py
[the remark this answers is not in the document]
sfriedowitz: No CLI changes in this PR, as we are planning to make broader interface changes down the line.

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
+# lm-buddy entrypoint to run evaluations using a Prometheus inference server
+# see https://github.com/kaistAI/prometheus/blob/main/evaluation/benchmark/run_absolute_scoring.py
```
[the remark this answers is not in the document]
sfriedowitz: True, if you are adding a module-level comments we should use docstrings as `"""`

src/lm_buddy/jobs/_entrypoints/prometheus.py
sfriedowitz: Either way, I would set this as a global variable at the top level of the file, e.g, `PROMETHEUS_MODEL_MAX_TOKENS = 3072`

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
...
+    ds = load_dataset("json", data_files=str(output_fname), split="train")
+    ds.save_to_disk(output_hf_name)
+
+    return str(output_hf_name)
+
+
+def run_prometheus(config: PrometheusJobConfig, artifact_loader: ArtifactLoader):
+    # load dataset from W&B artifact
+    hf_loader = HuggingFaceAssetLoader(artifact_loader)
+    data = hf_loader.load_dataset(config.dataset)
```
sfriedowitz, line 139: This line should be removed within the scope of the initialized W&B run, or else artifact lineages will not be connected. The tracking run must be active to link an input artifact (the dataset in this case).

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # get the tokenizer
+    tokenizer_config = AutoTokenizerConfig(load_from=config.prometheus.inference.engine)
+    tokenizer = hf_loader.load_pretrained_tokenizer(tokenizer_config)
```
sfriedowitz, line 143: This line should be removed within the scope of the initialized W&B run, or else artifact lineages will not be connected. The tracking run must be active to link an input artifact (the tokenizer in this case).

src/lm_buddy/jobs/_entrypoints/prometheus.py
sfriedowitz: Moving the dataset and tokenizer loading to within the `run_eval` function seems most appropriate, since that is only called after run initialization.

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
...
+            output_dataset_name = run_eval(config, data, tokenizer, client)
+
+            # store HF dataset as a directory artifact
+            artifact = build_directory_artifact(
+                dir_path=output_dataset_name,
+                artifact_name=config.tracking.name,
+                artifact_type=ArtifactType.DATASET,
+                reference=False,
+            )
+            print("[i] Logging artifact for evaluation results...")
```
sfriedowitz, line 161:
```suggestion
            print("Logging artifact for evaluation results...")
```

src/lm_buddy/jobs/configs/prometheus.py
```diff
@@ -0,0 +1,35 @@
...
+    max_score: int = 5
+    enable_tqdm: bool = False
+    tmp_folder: str = "/tmp"
+
+
+class PrometheusJobConfig(LMBuddyJobConfig):
+    """Configuration to run a prometheus job."""
+
+    dataset: TextDatasetConfig = Field(
+        description="dataset (json artifact from which we'll extract `text_field`)"
```
sfriedowitz, line 25:
```suggestion
        description="Dataset of text completions to evaluate using the Prometheus judge model."
```

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
...
+                result["prometheus_output"].append(feedback)
+                result["prometheus_score"].append(score)
+
+            # dump sample results incrementally
+            file.write(json.dumps(result) + "\n")
+
+    # convert plain json dataset in HF format
+    output_hf_name = str(Path(config.evaluation.tmp_folder) / "hf" / tracking_name)
+    ds = load_dataset("json", data_files=str(output_fname), split="train")
+    ds.save_to_disk(output_hf_name)
```
veekaybee, line 131: question on where these currently get persisted: are they saved to disk on the head node in Ray? If so, are we ok with this file being there rather than in wandb?

src/lm_buddy/jobs/_entrypoints/prometheus.py
aittalam: Retries include both openai calls and result parsing, as this could be malformed for many different reasons.
I will factor both of them out!

src/lm_buddy/integrations/vllm.py
aittalam: Yes, this is why I wanted to have a vLLM-specific configuration, as the parameters can be way more than those which are prometheus-specific.
Also, if we ever decide to support other types of inference service (e.g. TGI which is what the original Prometheus was built on) which are not API-compatible we can just have a new config for them without having to tamper with the prometheus-specific params.
TL/DR: yes, I will add the URL for reference, that's a great idea!

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,165 @@
...
+        raise BadResponseError(f"Server returned a malformed response ({e})", e)
+
+    return feedback, score
+
+
+def instruction_to_prompt(instruction: str) -> str:
+    """Given some text containing Prometheus instructions, transform it
+    into an actual llama-2 prompt.
+    """
+    conv = get_conv_template("llama-2")
```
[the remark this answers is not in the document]
aittalam: I see your point... In theory yes, it would be cleaner than having something hard-coded in our entrypoint.
In practice, I think that might be overkill as we are developing for a model that has been trained to work in a very specific way and we cannot forsee if these things will change or not in future versions (or if anything else will).
As the extra effort is low I will make these configurable anyway, but I will add defaults in the config class so they do not have to be explicitly specified.

src/lm_buddy/jobs/_entrypoints/prometheus.py
aittalam: Good catch, I fixed it and now have lineage working:
image

src/lm_buddy/jobs/_entrypoints/prometheus.py
aittalam: My rationale is:
- in the ray+wandb scenario, this is persisted locally and then synced to wandb (the file on the pod can then be discarded)
- in the local+wandb scenario, this is persisted locally and then synced to wandb (the local file can then be discarded if desired, which is why the default directory is /tmp)
- in the local+no_tracking scenario, this is just saved locally 
- note that we can have tracking or not on ray and save the datasets to some persisted folders, which is why I allowed users to specify an output folder.

I realised though that what I called `tmp_folder` should be called `output_folder` instead so that it is more clear it can be non-temporary.

We should probably also provide the possibility to choose in the config whether we want to save a reference artifact or the whole file. Ideally it should be part of `tracking` but as it involves all the other entrypoints I think it deserves a separate discussion

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,187 @@
...
+    # Register a dataset file artifact if tracking is enabled
+    if config.tracking:
+        with wandb_init_from_config(config.tracking, job_type=LMBuddyJobType.EVALUATION):
+            # load dataset from W&B artifact
+            hf_loader = HuggingFaceAssetLoader(artifact_loader)
+            data = hf_loader.load_dataset(config.dataset)
+
+            # get the tokenizer
+            tokenizer_config = AutoTokenizerConfig(load_from=config.prometheus.inference.engine)
+            tokenizer = hf_loader.load_pretrained_tokenizer(tokenizer_config)
```
sfriedowitz, line 163: Moving this into the `run_eval` method would reduce some code duplication, although not a huge issue.

src/lm_buddy/jobs/_entrypoints/prometheus.py
sfriedowitz: I think the discussion of persistent disk storage locations is going to be a bigger thing to address in the future. For instance, if we run an entire workflow at once with the new API I am developing, I think it makes sense to have a single folder that houses all the generated artifacts from that run. This will likely involve attributing UIDs of some kind to the workflow as a whole, and to the individual tasks within a workflow.

That being said, I'm not sure what thte best thing to do for now is. We don't really have set storage locations for where this code writes to disk at. The only thing that exists is the `RAY_STORAGE_PATH` env var on our clusters, which exclusively controls the output locations for invocations of Ray train during finetuning.

I think for now, the best we can do is have the user specify on the config an output location. But we will certainly end up revisiting this as the library evolves.

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -0,0 +1,187 @@
...
+        # load dataset from W&B artifact
+        hf_loader = HuggingFaceAssetLoader(artifact_loader)
+        data = hf_loader.load_dataset(config.dataset)
+
+        # get the tokenizer
+        tokenizer_config = AutoTokenizerConfig(load_from=config.prometheus.inference.engine)
+        tokenizer = hf_loader.load_pretrained_tokenizer(tokenizer_config)
+
+        output_dataset_name = run_eval(config, data, tokenizer, client)
+        print(f"[i] Evaluation results stored in {output_dataset_name}")
```
sfriedowitz, line 187:
```suggestion
        print(f"Evaluation results stored in {output_dataset_name}")
```

sfriedowitz (approved):
I'll approve this for now as most comments have been addressed. I left a few other smaller nit comments you may want to tackle as well in this PR.

In a followup, I think it would be good to take another iteration at the following:
- Refining how we store generated results to disk, and the format of the W&B artifacts. E.g., do we log cumulative metrics from the eval as a Run Table for easy visualization, in addition to the file artifact?
- Adding test coverage, both for ser/de issues on the new configs, and for the job entrypoint as a whole.

src/lm_buddy/jobs/_entrypoints/prometheus.py
veekaybee: Sounds good, let's merge and file a separate issue for this to keep going on the storage in a new ticket

veekaybee (approved):
🚀

Issue #49, opened 2024-02-13
Warning on `trust_remote_code` on datasets


aittalam:
Running lm-harness with the following config:

```
# Model to evaluate
model:
  load_from: "tiiuae/falcon-7b"
  torch_dtype: "bfloat16"

# Settings specific to lm_harness.evaluate
evaluator:
  tasks: ["hellaswag", "mmlu"]
  num_fewshot: 5
  limit: 10

quantization:
  load_in_4bit: True
  bnb_4bit_quant_type: "fp4"
  bnb_4bit_compute_dtype: "bfloat16"

# Tracking info for where to log the run results
tracking:
  name: "tiiuae-falcon-7b-trust"
  project: "davide-testing-flamingo"
  entity: "mozilla-ai"
```

Returns the following warnings:

```
53 tmp/ray/session_2024-02-01_09-08-37_813595_7/runtime_resources/pip/1989896a12212ac88dcdfee7f5a5fef9a4046e6c/virtualenv/lib/python3.10/site-packages/datasets/load.py:1429: FutureWarning: The repository for hellaswag contains custom code which must be executed to correctly load the dataset. You can inspect the repository content at https://hf.co/datasets/hellaswag
54 You can avoid this message in future by passing the argument `trust_remote_code=True`.
55 Passing `trust_remote_code=True` will be mandatory to load this dataset from the next major release of `datasets`.
56 ...
57 /tmp/ray/session_2024-02-01_09-08-37_813595_7/runtime_resources/pip/1989896a12212ac88dcdfee7f5a5fef9a4046e6c/virtualenv/lib/python3.10/site-packages/datasets/load.py:1429: FutureWarning: The repository for hails/mmlu_no_train contains custom code which must be executed to correctly load the dataset. You can inspect the repository content at https://hf.co/datasets/hails/mmlu_no_train
58 You can avoid this message in future by passing the argument `trust_remote_code=True`.
59 Passing `trust_remote_code=True` will be mandatory to load this dataset from the next major release of `datasets`.
```

Example: http://10.145.91.219:8265/#/jobs/raysubmit_ap35cU8CVdhAtDaa

The problem does not disappear when `trust_remote_code` is added to the model section. It has to be added to `load_dataset` as shown [here](https://huggingface.co/docs/datasets/loading#local-loading-script). This is not our problem rn but it might impact us at some point.

veekaybee:
This is something I can look into from the lm-harness side

[...]

PR #76, opened 2024-03-05
Pinning lm-harness at the git hash

veekaybee:
## What's changing

We'd like to pin lm-harness to the hash resolving trust_remote_code (see issue here:https://github.com/mozilla-ai/lm-buddy/issues/49), and also updating to include wandb for logging


sfriedowitz (commented):
Can you update the contributing and readme to see if there is mention of UV? We should remove that for now.

Also a patch version bump.

sfriedowitz (approved):

PR #77, opened 2024-03-11
lm buddy ragas draft

imtihan:
## What's changing
Added ragas entrypoint. Still need to run a full test on it but it code is generally ready for review. 
## Related Jira Ticket
https://mzai.atlassian.net/browse/RD2024-5


src/lm_buddy/jobs/ragas/entrypoint.py
```diff
@@ -0,0 +1,110 @@
+import os
```
veekaybee, line 1: i wonder if we should have /jobs/eval/ragas...etc since we will have more than 1 eval type as a structure

src/lm_buddy/jobs/ragas/ragas_config.py
```diff
@@ -0,0 +1,99 @@
...
+        answer_relevancy,
+        context_recall,
+        context_precision,
+    ]
+
+    # path to store the generated ratings/evaluations of each dataset sample
+    result_path: str | Path | None = None
+
+
+class RagasvLLMJudgeConfig(BaseLMBuddyConfig):
```
veekaybee, line 32: is this specific to vLLM, or can it take any hosted server? if it can, we can make this more general than vLLMJudge

imtihan: Ragas specifies a bunch of endpoints that are openai api compatible, makes sense to change it to openai api endpoint since that's more precise

src/lm_buddy/jobs/_entrypoints/ragas.py
```diff
@@ -0,0 +1,105 @@
...
+from lm_buddy.integrations.wandb.artifact_utils import (
+    ArtifactType,
+    build_directory_artifact,
+    default_artifact_name,
+)
+from lm_buddy.integrations.wandb.run_utils import wandb_init_from_config
+from lm_buddy.jobs.common import EvaluationResult, LMBuddyJobType
+from lm_buddy.jobs.configs import RagasJobConfig
+
+RAGAS_METRICS_MAP = {
```
veekaybee, line 21: could we get these from `RagasEvaluationConfig`?

src/lm_buddy/jobs/configs/ragas.py
```diff
@@ -0,0 +1,60 @@
...
+    The dataset must be formatted in the RAG context, with the question, generated answer,
+    the contexts (retrieved), and optionally a ground truth field.
+    """
+
+    # vllm inference server for generation
+    judge: VLLMCompletionsConfig = Field(description="Externally hosted Ragas judge model.")
+
+    # dataset containing the relevant fields required for ragas evaluation
+    dataset: TextDatasetConfig = Field(
+        description="Dataset of text completions to evaluate using the Prometheus judge model."
```
veekaybee, line 50: should this be `Ragas`?

src/lm_buddy/jobs/_entrypoints/ragas.py
```diff
@@ -0,0 +1,105 @@
...
+    # convert plain json dataset in HF format
+    output_dataset_path = Path(config.evaluation.output_folder) / "hf" / tracking_name
+    ds = load_dataset("json", data_files=str(output_fname), split="train")
+    ds.save_to_disk(output_dataset_path)
+
+    return output_dataset_path
+
+
+def run_ragas(config: RagasJobConfig, artifact_loader: artifact_loader) -> EvaluationResult:
+    # Run eval and store output in local filename
```
veekaybee, line 72: might want to change the comment since we also have a `run_eval` function, to be clearer on what this does

src/lm_buddy/jobs/_entrypoints/ragas.py
```diff
@@ -0,0 +1,105 @@
...
+
+            # Create a directory artifact for the HF dataset
+            dataset_artifact = build_directory_artifact(
+                dir_path=output_dataset_path,
+                artifact_name=default_artifact_name(run.name, artifact_type=ArtifactType.DATASET),
+                artifact_type=ArtifactType.DATASET,
+                reference=False,
+            )
+
+            print("Logging artifact for evaluation dataset...")
```
veekaybee, line 85: make logs clearer that this is running ragas?

examples/configs/ragas/ragas_config.yaml
```diff
@@ -0,0 +1,38 @@
...
+evaluation:
+  #choose which metrics to report from https://docs.ragas.io/en/latest/concepts/metrics/index.html
+  metrics: [
+        "faithfulness",
+        "answer_relevancy",
+        "context_recall",
+        "context_precision",
+    ]
+  # dummy openAI api key, needed for ragas shenanigans
+  openai_api_key: "nokey"
```
veekaybee, line 28: we can set to "EMPTY" following lm-harness

examples/configs/ragas/ragas_config.yaml
```diff
@@ -0,0 +1,38 @@
...
+  # any embedding model in huggingface, will be loaded in-memory (typically <1gb)
+  embedding_model: "sentence-transformers/all-mpnet-base-v2"
+
+tracking:
+  name: "lm-buddy-ragas"
+  project: "lm-buddy-examples"
+  entity: "mozilla-ai"
```
veekaybee, line 35: let's change this to a "sample" entity since the code will be public

src/lm_buddy/jobs/_entrypoints/ragas.py
```diff
@@ -0,0 +1,105 @@
...
+        model=config.ragas_inference_server.engine,
+        openai_api_key=config.evaluation.openai_api_key,
+        # get api endpoint from environment variable
+        openai_api_base=config.judge.inference.base_url,
+        max_tokens=config.judge.max_tokens,
+        temperature=config.judge.temperature,
+        top_k=config.judge.top_k,
+    )
+
+    result = evaluate(
```
veekaybee, line 51: it might be helpful to either import ragas.evaluate directly or just add a comment for clarity that this is running RAGAS evaluation method since we have other evaluation methods in this class

veekaybee:
Would also be helpful to write some documentation here under evaluation_concepts: https://github.com/mozilla-ai/lm-buddy/tree/main/docs/source

src/lm_buddy/jobs/_entrypoints/ragas.py
```diff
@@ -0,0 +1,105 @@
...
+    evaluation_dataset = hf_loader.load_dataset(config.dataset)
+
+    # ragas custom model args
+    ragas_args = {}
+
+    # load embedding model
+    ragas_args["embeddings"] = HuggingFaceEmbeddings(model_name=config.evaluation.embedding_model)
+
+    # configure ragas to point to vllm instance for generation
+    ragas_args["llm"] = ChatOpenAI(
```
veekaybee, line 41: Are we able to bypass langchain here and use the vllm implementation since we already have access? https://github.com/vllm-project/vllm/blob/f721096d48a7e3b98dffcb9b400bf58989cef64d/examples/openai_completion_client.py#L7. worth checking https://api.python.langchain.com/en/latest/language_models/langchain_core.language_models.base.BaseLanguageModel.html

examples/configs/ragas/ragas_config.yaml
sfriedowitz: Do we need this on the config? If running on Ray, a user can inject this into their workers by the `runtime_env` attribute.

We aren't passing other keys (W&B, HF) as part of the configs, we are using env vars.

pyproject.toml
```diff
@@ -4,12 +4,13 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "lm-buddy"
-version = "0.4.0"
+version = "0.4.1"
```
sfriedowitz, line 7:
```suggestion
version = "0.5.0"
```

Since its a new feature, not a patch.

src/lm_buddy/jobs/configs/ragas.py
```diff
@@ -0,0 +1,60 @@
...
+from lm_buddy.integrations.huggingface.dataset_config import TextDatasetConfig
+from lm_buddy.integrations.vllm import VLLMCompletionsConfig
+from lm_buddy.integrations.wandb import WandbRunConfig
+from lm_buddy.types import BaseLMBuddyConfig
+
+
+class RagasEvaluationConfig(BaseLMBuddyConfig):
+    """Parameters specifically required for RAGAs Evaluation"""
+
+    metrics: list[str] | None = [
```
sfriedowitz, line 13: If you want to ensure the values passed here are correct, Pydantic validation can help.

```
from typing import Literal

RagasEvaluationMetric = Literal["faithfulness", "answer_relevancy", ...]

class RagasEvaluationConfig(BaseLMBuddyConfig):
    metrics: list[RagasEvaluationMetric]
```

Also, [Python has funky behavior with default values for mutable data structures.](https://docs.python-guide.org/writing/gotchas/#:~:text=Mutable%20Default%20Arguments%C2%B6). If you want to specify a default here, you should use the `Field(..., default_factory=list())` function from Pydantic.

sfriedowitz (commented):
overall looking pretty good, a few comments for some gotchas.

src/lm_buddy/jobs/configs/ragas.py
imtihan: oh lol yeah

examples/configs/ragas/ragas_config.yaml
imtihan: unfortunately it's an artifact of Ragas, you need a dummy key when hitting non openai endpoints

examples/configs/ragas/ragas_config.yaml
aittalam, line 1: We have an `examples/configs/evaluation` folder, wdyt about putting all the configs there?
I just realised I agreed it with that but haven't done it for prometheus yet, if that's ok for you feel free to move both config files to the same `evaluation` directory

examples/configs/ragas/ragas_config.yaml
aittalam: Same for prometheus' openai client! I solved it by just adding the following to my `~/.zshrc` file:
```
export OPENAI_API_KEY="Ceci n'est pas une openai key"
```
... does that work in your case?

examples/configs/ragas/ragas_config.yaml
aittalam: Good point. We have the same for prometheus and lm-harness, would you mind updating them too pls?

src/lm_buddy/jobs/_entrypoints/ragas.py
aittalam: We do have a (barebones) OpenAI client implementation in the prometheus entrypoint, perhaps it might help if we moved it to a common library and reuse it?

src/lm_buddy/jobs/_entrypoints/ragas.py
```diff
@@ -0,0 +1,105 @@
...
+        model=config.ragas_inference_server.engine,
+        openai_api_key=config.evaluation.openai_api_key,
```
aittalam, line 43: If the env variable approach works with langchain's implementation of the openai client, you might not need this line

examples/configs/ragas/ragas_config.yaml
sfriedowitz: Yes if its just a dummy key, can we either set it as an env var or just hard-code a dummy for the client? it seems inapprorpiate to pass the dummy via the config since its not _really_ a user-controlled parameter.

imtihan: I like the idea of hardcoding it in. If we want to support actually using OpenAI models later we can add that in

src/lm_buddy/jobs/_entrypoints/ragas.py
imtihan: we don't want to import the jobs directly into the config files, creates issues with pydantic apparently. Implemented it this way as a suggestion from @sfriedowitz

src/lm_buddy/jobs/_entrypoints/ragas.py
imtihan: added alias to evaluate import `from ragas import evaluate as ragas_evaluate
` so that there is no confusion

src/lm_buddy/jobs/_entrypoints/ragas.py
sfriedowitz: The metrics imported from `ragas` are functions. They cannot be serialized. The appropriate metric function must then be looked up by string identifier, which Imtihan has implemented here.

src/lm_buddy/jobs/_entrypoints/ragas.py
imtihan: I'm looking into how ragas is calling the completions in each evaluate method to see if it will work

examples/configs/ragas/ragas_config.yaml
imtihan: Simply removing the key from the yaml file passes the default, so I'll just remove it from the config yaml.

src/lm_buddy/jobs/_entrypoints/ragas.py
imtihan: It looks like ragas basically uses langchain under the hood to load the models https://github.com/explodinggradients/ragas/blob/962d40d1f5f45fe57266ec35176329b07dfc0ae5/src/ragas/adaptation.py#L3, and it doesn't have direct overrides for vllm (supporting it essentially through langchain) so I think we will have to keep this dependency for now. They might add native support for it later, or we can contribute back to it and then refactor this out after the fact if it makes sense.

veekaybee:
Thanks for this @imtihan ! LGTM once tests pass 🚀

veekaybee (approved):

PR #78, opened 2024-03-11
Bumping the library version and pinning lm-eval-harness to a PyPI Distribution

veekaybee:
## What's changing

We wanted to pin lm-eval to a git hash to encompass the latest changes, but PyPI doesn't allow to publish packages with git-hash pinned dependencies, so reverting for now so we can update the PyPI version with the latest Ray on publish. 

See errors here: https://github.com/mozilla-ai/lm-buddy/actions/runs/8238935806/job/22531036074

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "lm-buddy"
-version = "0.2.7"
+version = "0.2.8"
 authors = [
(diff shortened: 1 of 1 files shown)


sfriedowitz (approved):
Thanks for doing this 🙇

sfriedowitz:
(we could in theory switch back to `uv` in our CI, but no worries either way)

veekaybee:
Yup, will do once we bump that lm-harness version

PR #84, opened 2024-03-14
Initial LMBuddy class for running jobs

sfriedowitz:
## What's changing

- Removes the `run_job` method in favor of a class `LMBuddy` that has methods for `finetune` and `evaluate`
- Implements job-result data structures for returning data generated by job entrypoints, instead of simply having the entrypoints write to disk/W&B
- Implements a new `LoadableAssetPath` type and associated data structures to represent any `load_from` path for a HF asset. See inline comments for motivation for this change.

Note that the CLI API is not changed by these internal changes, so you can still execute the package as a Ray entrypoint in the same manner as before.

## How to test it

- Run the test suite
- Pull the branch and try running some jobs with the new interface

In follow-up PRs into this dev branch, I would like to do the following:
- https://mzai.atlassian.net/browse/RD2024-153
- https://mzai.atlassian.net/browse/RD2024-152


src/lm_buddy/integrations/huggingface/dataset_config.py
```diff
@@ -10,25 +9,19 @@
 class DatasetConfig(BaseLMBuddyConfig):
     """Base configuration to load a HuggingFace dataset."""
 
-    load_from: HuggingFaceRepoConfig | WandbArtifactConfig
+    load_from: LoadableAssetPath
     split: str | None = None
     test_size: float | None = None
     seed: int | None = None
 
-    _validate_load_from_string = field_validator("load_from", mode="before")(
```
sfriedowitz, line 18: These validators are built into the `LoadableAssetPath` type now

src/lm_buddy/integrations/wandb/artifact_utils.py
```diff
@@ -91,57 +91,21 @@ def build_directory_artifact(
...
     """
     artifact = wandb.Artifact(artifact_name, type=artifact_type)
     for table_name, table_data in tables.items():
-        table = wandb.Table(data=table_data, columns=columns)
+        table = wandb.Table(data=table_data)
         artifact.add(table, name=table_name)
     return artifact
-
-
-def build_file_artifact(
```
sfriedowitz, line 115: This was not being used

src/lm_buddy/buddy.py
```diff
@@ -0,0 +1,41 @@
...
+from lm_buddy.jobs.configs import (
+    EvaluationJobConfig,
+    FinetuningJobConfig,
+    LMHarnessJobConfig,
+    PrometheusJobConfig,
+    SimpleJobConfig,
+)
+
+
+class LMBuddy:
```
sfriedowitz, line 13: Per our discussions over Zoom, we decided that it would be difficult to built a single workflow that runs all tasks consecutively at the moment. However, I do feel it is nicer to have this class that can dispatch to the various jobs, even if it doesn't support running things in a pipeline atm.

So for now, I am adding this as a thin wrapper around the entrypoint functions, and returning explicit return types for the data generated by the jobs.

examples/configs/evaluation/lm_harness_hf_config.yaml
```diff
@@ -4,7 +4,7 @@ model:
   torch_dtype: "bfloat16"
 
 # Settings specific to lm_harness.evaluate
-evaluator:
+evaluation:
```
sfriedowitz, line 7: This was changed to be consistent with the prometheus config, which names it `evaluation` instead of `evaluator`

examples/configs/evaluation/lm_harness_quantized_config.yaml
sfriedowitz, line 1: This is a duplicate of the other `hf_config.yaml` file since it already has the quantization section specified.

src/lm_buddy/integrations/huggingface/asset_loader.py
```diff
@@ -66,23 +63,18 @@ class HuggingFaceAssetLoader:
...
         self._artifact_loader = artifact_loader
 
-    def resolve_asset_path(self, path: HuggingFaceAssetPath) -> tuple[str, str | None]:
-        """Resolve the actual HuggingFace name/path from a config.
-
-        Currently, two config types contain references to a loadable HuggingFace path:
-        (1) A `HuggingFaceRepoConfig` that contains the repo path directly
-        (2) A `WandbArtifactConfig` where the filesystem path is resolved from the artifact
-        """
+    def resolve_asset_path(self, path: LoadableAssetPath) -> str:
```
sfriedowitz, line 66: Note that I've changed the signature of this method to only return a string path, rather than a string path AND a `revision` field.

I felt that included `revision` within the path resolution logic was a poor abstraction for the following reasons:
- It is only applicable to HF Hub repos, not artifacts or direct filesystem paths
- We aren't even really using it

If we want to add back the ability to load from a specific HF Hub repo revision, we can include that parameter as part of a separate config (e.g., on the `AutoModelConfig` directly).

src/lm_buddy/jobs/_entrypoints/finetuning.py
```diff
@@ -101,14 +106,31 @@ def training_function(config_data: dict[str, Any]):
...
-        with wandb_init_from_config(config.tracking, resume=WandbResumeMode.MUST) as run:
-            model_artifact = build_directory_artifact(
-                artifact_name=default_artifact_name(run.name, ArtifactType.MODEL),
-                artifact_type=ArtifactType.MODEL,
-                dir_path=f"{result.checkpoint.path}/{RayTrainReportCallback.CHECKPOINT_NAME}",
-                reference=True,
-            )
-            print("Logging artifact for model checkpoint...")
-            artifact_loader.log_artifact(model_artifact)
+    ckpt_path, artifact_config = None, None
```
sfriedowitz, line 109: In a follow-up PR (https://mzai.atlassian.net/browse/RD2024-152), I would like to refactor a bit how we are generating artifacts and results in these methods. 

The issue is that because the tracking field is optional, we repeatedly end up having to write two code branches, (1) for when we initialize a W&B run and create an artifact and (2) for when we run the job without tracking or artifacts. It's almost like we want something like `maybe_initialize_wandb_run` that handles the optionality of the tracking service.

src/lm_buddy/jobs/_entrypoints/lm_harness.py
```diff
@@ -1,43 +1,45 @@
...
 from lm_buddy.jobs.configs import LMHarnessJobConfig, LocalChatCompletionsConfig
+from lm_buddy.paths import LoadableAssetPath
 
 
-def get_numeric_metrics(
+def get_per_task_dataframes(
     results: dict[str, dict[str, Any]],
-) -> dict[str, list[tuple[str, float]]]:
-    """Filter non-numeric values from the evaluation results.
+) -> dict[str, pd.DataFrame]:
```
sfriedowitz, line 30: Returning these as `pd.DataFrame` now since the type is a bit cleaner to work with when included on the `EvaluationResult` class.

src/lm_buddy/buddy.py
```diff
@@ -0,0 +1,41 @@
...
+    """Your buddy in the (L)LM space.
+
+    Simple wrapper around executable functions for tasks available in the library.
+    """
+
+    def __init__(self, artifact_loader: ArtifactLoader = WandbArtifactLoader()):
+        self._artifact_loader = artifact_loader
+
+    def simple(self, config: SimpleJobConfig) -> SimpleResult:
```
sfriedowitz, line 22: Should I just delete the `simple` example? It's kinda irrelevant at this point, and doesn't fit in semantically here very well.

veekaybee:
Looking now! Pulled the branch, reading the instructions for `direct_job_execution.ipynb` and have put together a a script like this. Where do we specify the cluster information in the new workflow?  Seems like it should be somewhere here right? `FinetuningRayConfig`

```python
from ray.job_submission import JobSubmissionClient
from pathlib import Path


from lm_buddy import LMBuddy
from lm_buddy.jobs.configs import (
    FinetuningJobConfig,
    FinetuningRayConfig,
    LMHarnessJobConfig,
    LMHarnessEvaluationConfig,
)
from lm_buddy.integrations.huggingface import (
    AutoModelConfig,
    TextDatasetConfig,
    TrainerConfig,
    AdapterConfig,
)
from lm_buddy.integrations.wandb import WandbRunConfig

# Base model to finetune from HuggingFace
model_config = AutoModelConfig(load_from="distilgpt2")

# Text dataset for finetuning
dataset_config = TextDatasetConfig(
    load_from="imdb",
    split="train[:100]",
    text_field="text",
)

# HuggingFace trainer arguments
trainer_config = TrainerConfig(
    max_seq_length=256,
    per_device_train_batch_size=8,
    learning_rate=1e-4,
    num_train_epochs=1,
    logging_strategy="steps",
    logging_steps=1,
    save_strategy="epoch",
    save_steps=1,
)

# LORA adapter settings
adapter_config = AdapterConfig(
    peft_type="LORA",
    task_type="CAUSAL_LM",
    r=8,
    lora_alpha=16,
    lora_dropout=0.2,
)

# Define tracking for finetuning run
tracking_config = WandbRunConfig(
    name="example-finetuning",
    project="lm-buddy-examples",  # Update to your project name
    entity="mozilla-ai",  # Update to your entity name
)

# Ray train settings
ray_config = FinetuningRayConfig(
    use_gpu=False,  # Change to True if GPUs are available on your machine
    num_workers=2,
)

# Full finetuning config
finetuning_config = FinetuningJobConfig(
    model=model_config,
    dataset=dataset_config,
    trainer=trainer_config,
    adapter=adapter_config,
    tracking=tracking_config,
    ray=ray_config,
)
```

sfriedowitz:
>Where do we specify the cluster information in the new workflow?

Nothing is changing in how you specify the cluster information. The CLI of the package is not changed, so you can use the same commands as an entrypoint to a Ray job submission using their SDK.

src/lm_buddy/buddy.py
veekaybee: Yeah, let's delete

src/lm_buddy/buddy.py
```diff
@@ -0,0 +1,41 @@
...
+        """Run a simple test task."""
+        return run_simple(config)
+
+    def finetune(self, config: FinetuningJobConfig) -> FinetuningResult:
+        """Run a supervised finetuning task with the provided configuration."""
+        return run_finetuning(config, self._artifact_loader)
```
veekaybee, line 28: This is kind of confusing because we run_finetuning but we return results from run_finetuning, could be as simple as assigning it for legibility?

```python
def finetune(self, config: FinetuningJobConfig) -> FinetuningResult:
        """Run a supervised finetuning task with the provided configuration."""
        finetuning_results = run_finetuning(config, self._artifact_loader)
        return finetuning_results
```

src/lm_buddy/cli/run.py
```diff
@@ -22,25 +23,25 @@ def group():
 @click.option("--config", type=str)
 def run_simple(config: str) -> None:
     config = SimpleJobConfig.from_yaml_file(config)
-    lm_buddy.run_job(config)
+    buddy.simple(config)
```
veekaybee, line 26: we can remove

src/lm_buddy/buddy.py
```diff
@@ -0,0 +1,41 @@
...
+)
+
+
+class LMBuddy:
+    """Your buddy in the (L)LM space.
+
+    Simple wrapper around executable functions for tasks available in the library.
+    """
+
+    def __init__(self, artifact_loader: ArtifactLoader = WandbArtifactLoader()):
```
veekaybee, line 19: I wonder..if Artifact is a WandB specific term, does it make sense to call the more abstract loader a ModelLoader or similar for clarity?

src/lm_buddy/integrations/huggingface/asset_loader.py
```diff
@@ -66,23 +63,18 @@ class HuggingFaceAssetLoader:
...
+        """Resolve the actual HuggingFace name/path from a `LoadableAssetPath`."""
         match path:
-            case HuggingFaceRepoConfig(repo_id, revision):
-                load_path, revision = repo_id, revision
+            case FilesystemPath(value):
```
veekaybee, line 69: suggestion: just FilePath

src/lm_buddy/integrations/huggingface/tokenizer_config.py
```diff
@@ -1,17 +1,10 @@
...
-from lm_buddy.integrations.wandb import WandbArtifactConfig
+from lm_buddy.paths import LoadableAssetPath
 from lm_buddy.types import BaseLMBuddyConfig
 
 
 class AutoTokenizerConfig(BaseLMBuddyConfig):
     """Settings passed to a HuggingFace AutoTokenizer instantiation."""
 
-    load_from: HuggingFaceRepoConfig | WandbArtifactConfig
+    load_from: LoadableAssetPath
```
veekaybee, line 8: maybe also just `AssetPath`? or `ModelPath`

src/lm_buddy/integrations/wandb/artifact_utils.py
```diff
@@ -91,57 +91,21 @@ def build_directory_artifact(
...
+    tables: dict[str, pd.DataFrame],
 ) -> wandb.Artifact:
     """Build an artifact containing one or more table entries.
 
     Args:
         artifact_name (str): Name of the artifact.
         artifact_type (ArtifactType): Type of artifact.
-        columns (list[str]): Column names for the tables.
-        tables (dict[str, list[list[Any]]]): Mapping from table name to table rows.
+        tables (dict[str, pd.DataFrame]): Mapping from table name to table data
```
veekaybee, line 101: did pandas end up being easier than the previous implementation?

src/lm_buddy/jobs/_entrypoints/lm_harness.py
```diff
@@ -1,43 +1,45 @@
...
+    """Create a `pd.DataFrame` of numeric metrics for each evaluation task.
 
     This is necessary because artifact tables must have a single datatype for each column.
 
     lm-harness returns mostly numeric values, but there are also some misc string-valued metrics.
     Filtering down to only numeric values allows us to produce a valid table artifact.
     """
-    numeric_results = {}
-    for key, data in results.items():
+    dfs = {}
```
veekaybee, line 38: maybe something more descriptive like `dataframe_results`?

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -138,7 +139,7 @@ def run_eval(
             result["prometheus_output"] = []
             result["prometheus_score"] = []
 
-            for idx in range(config.evaluation.num_answers):
+            for _ in range(config.evaluation.num_answers):
```
veekaybee, line 142: nit/suggestion, change to `i` for clarity

src/lm_buddy/jobs/_entrypoints/prometheus.py
```diff
@@ -149,32 +150,50 @@ def run_eval(
...
-
-            # store HF dataset as a directory artifact
-            artifact = build_directory_artifact(
-                dir_path=output_dataset_name,
-                artifact_name=config.tracking.name,
+        with wandb_init_from_config(config.tracking, job_type=LMBuddyJobType.EVALUATION) as run:
+            output_dataset_path = run_eval(config, artifact_loader, client)
+
+            # Create a directory artifact for the HF dataset
+            dataset_artifact = build_directory_artifact(
```
veekaybee, line 173: suggestion: might be helpful to add types to any variable assignments where we generate artifacts or data outputs so we can easily tell what's generated

src/lm_buddy/jobs/_entrypoints/simple.py
```diff
@@ -1,5 +1,6 @@
 import ray
```
veekaybee, line 1: i think we can remove this also

src/lm_buddy/paths.py
```diff
@@ -0,0 +1,68 @@
...
+
+from huggingface_hub.utils import HFValidationError, validate_repo_id
+from pydantic import BeforeValidator
+
+from lm_buddy.integrations.wandb import WandbArtifactConfig
+from lm_buddy.types import BaseLMBuddyConfig
+
+
+class FilesystemPath(BaseLMBuddyConfig):
+    """Absolute path to an object on the filesystem"""
```
veekaybee, line 12: suggestion: "on the local Linux filesystem of the head node of Ray" or something more specific about where results are being processed

src/lm_buddy/paths.py
```diff
@@ -0,0 +1,68 @@
...
+        s: string to test.
+    """
+    try:
+        validate_repo_id(s)
+        return True
+    except HFValidationError:
+        return False
+
+
+def validate_loadable_path(x: Any) -> Any:
```
veekaybee, line 42: suggestion `validate_path`

veekaybee:
Tests and left some comments, unit tests pass and sample job works! http://10.145.107.9:8265/#/jobs/raysubmit_85v8DpKbEVeKtbSB

sfriedowitz:
Thanks! Im a bit side tracked atm but will address most all of them in the next few hours.

src/lm_buddy/buddy.py
sfriedowitz: Artifact is a general term used by the 3 main tracking services (WandB, MlFlow, Comet). But I do think we will need to iterate on the purpose of this class in the near future.

I was looking at how other frameworks (e.g, HF `transformers`) handles tracking across multiple different tracking services. They implement a call-back based approach with something like a `WandbTrackingCallback` and `CometTrackingCallback`, etc..  I think we will want to abstract this interface to handle artifact/tracking stuff for the various frameworks eventually, but right now, I don't have a better option without a substantial bit of work.

src/lm_buddy/integrations/huggingface/tokenizer_config.py
sfriedowitz: It is more general than `ModelPath` since it applies to datasets and tokenizers, so I will call it `AssetPath`.

src/lm_buddy/integrations/wandb/artifact_utils.py
sfriedowitz: it did, yeah. We really just need a data structure for a generic tabular data. I was storing raw lists before, but `pd.DataFrame` was a bit easier to work with and it integrates automatically w/ W&B Tables.

src/lm_buddy/paths.py
sfriedowitz: This isn't always true because you may be running this code locally, in which case it simply corresponds to a path to load from disk.  I added the word "local" to clarify this.

src/lm_buddy/paths.py
sfriedowitz: Renamed to `validate_asset_path` in accordance with teh change in type name

veekaybee (approved):
Thanks for adressing! LGTM

Issue #79, opened 2024-03-11
Reinstate uv


veekaybee:
We'd like to use `uv` for pip dependencies but replaced it due to dependencies being unavailable as a git hash. This would re-implement it by reverting https://github.com/mozilla-ai/lm-buddy/pull/76

veekaybee:
Addressed here: https://github.com/mozilla-ai/lm-buddy/pull/86

PR #86, opened 2024-03-18
Reinstate uv for tests

veekaybee:
## What's changing

We can use UV again now that we don't use git-hash specified dependencies. See issue: https://github.com/mozilla-ai/lm-buddy/issues/79


sfriedowitz (approved):
GLTM

PR #87, opened 2024-03-18
Change CLI groups to finetune and evaluate

sfriedowitz:


src/lm_buddy/cli/evaluate.py
```diff
@@ -0,0 +1,27 @@
...
+from lm_buddy import LMBuddy
+from lm_buddy.jobs.configs import LMHarnessJobConfig, PrometheusJobConfig
+
+
+@click.group(name="evaluate", help="Run an LM Buddy evaluation job.")
+def group() -> None:
+    pass
+
+
+@group.command("lm-harness", help="Run the lm-harness evaluation job.")
```
sfriedowitz, line 12: Alternatively, I could try to implement a single CLI command called `lm-buddy evaluate` that accepts the underlying framework as a parameter. Thoughts?

veekaybee: what would be lowest lift? I think this is fine as-is for now

veekaybee (approved):
Looks good, thanks for this!

PR #88, opened 2024-03-18
RD2024-147: Implement LMBuddy class as generic job runner

sfriedowitz:
## What's changing

Combination of the two PRs:

1) https://github.com/mozilla-ai/lm-buddy/pull/84
2) https://github.com/mozilla-ai/lm-buddy/pull/87

Note that I have made a few breaking changes to the CLI and the `LMHarnessJobConfig` data structure here.
Also for @imtihan , the general entrypoint structure is the same for your Ragas work, but now returns an `EvaluationResult` return type that should be accounted for.


veekaybee (approved):
Approved since previous two PRs were reviewed and tested

PR #89, opened 2024-04-01
Add ability to pass a JSON string directly to CLI

sfriedowitz:

### DIFF
--- a/src/lm_buddy/cli/utils.py
+++ b/src/lm_buddy/cli/utils.py
@@ -1,0 +1,13 @@
+from pathlib import Path
+from typing import TypeVar
+
+from lm_buddy.jobs.configs.base import LMBuddyJobConfig
+
+ConfigType = TypeVar("ConfigType", bound=LMBuddyJobConfig)
+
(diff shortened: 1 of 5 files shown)


src/lm_buddy/cli/utils.py
```diff
@@ -0,0 +1,18 @@
+from pathlib import Path
+from typing import TypeVar
+
+from lm_buddy.jobs.configs.base import LMBuddyJobConfig
+
+ConfigType = TypeVar("ConfigType", bound=LMBuddyJobConfig)
+
+
+def parse_config_option(config_cls: type[ConfigType], config: str) -> ConfigType:
```
sfriedowitz, line 9: How does this work @imtihan ?

imtihan (approved):

src/lm_buddy/cli/utils.py
imtihan: like it, it's implicit

binaryaaron (approved):

PR #91, opened 2024-04-02
New mechanics for handling path types

sfriedowitz:


src/lm_buddy/integrations/wandb/artifact_utils.py
```diff
@@ -15,25 +16,13 @@ class ArtifactType(str, Enum):
     EVALUATION = "evaluation"
 
 
-class ArtifactURIScheme(str, Enum):
```
sfriedowitz, line 18: `PathPrefix` replaces the need for this

src/lm_buddy/jobs/_entrypoints/ragas.py
```diff
@@ -37,13 +36,14 @@ def run_eval(config: RagasJobConfig, artifact_loader: artifact_loader) -> Path:
     ragas_args = {}
 
     # load embedding model
```
sfriedowitz, line 38: @imtihan I think there were a few lingering bugs here that I've attempted to fix. The wrong types were being passed to the Langchain classes.

veekaybee (approved):
Looks good, thanks for adding! Worth thinking about: How much of this will we be able to port to the platform if we need to?

sfriedowitz:
>How much of this will we be able to port to the platform if we need to?

I think it will help with the platform a lot! For instance, being able to specify `hf://` vs `file://` vs `s3://` will likely be the difference between raw HF models and internal models that we have trained.

It's good groundwork for some of those goals IMO.

src/lm_buddy/paths.py
```diff
@@ -39,30 +36,41 @@ def is_valid_huggingface_repo_id(s: str):
...
+            raise ValueError(f"'{raw_path}' is not a valid HuggingFace repo ID.")
+    elif path.startswith(PathPrefix.WANDB):
+        # TODO: Validate the W&B path structure?
+        pass
+    else:
+        allowed_prefixes = {x.value for x in PathPrefix}
+        raise ValueError(f"'{path}' does not begin with an allowed prefix: {allowed_prefixes}")
+    return path
+
+
```
binaryaaron, line 55: might be worth a comment or longer docstring up top talking about the impl here

tests/utils.py
```diff
@@ -0,0 +1,52 @@
...
+from pydantic import BaseModel
+from wandb.sdk.artifacts.artifact_state import ArtifactState
+
+
+def copy_pydantic_json(model: BaseModel) -> BaseModel:
+    """Copy a Pydantic model through round-trip JSON serialization."""
+    return model.__class__.model_validate_json(model.model_dump_json())
+
+
+class FakeArtifactLoader:
```
binaryaaron, line 11: nit: move out of 'utils.py' to something more descriptive?

binaryaaron (approved):
minor notes; 🚢

src/lm_buddy/integrations/wandb/artifact_loader.py
```diff
@@ -31,14 +33,15 @@ class WandbArtifactLoader:
     This class makes external calls to the W&B API and is not suitable for test environments.
     """
 
-    def use_artifact(self, config: WandbArtifactConfig) -> wandb.Artifact:
+    def use_artifact(self, artifact_path: str) -> wandb.Artifact:
```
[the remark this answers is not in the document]
sfriedowitz: Nope that type is correct there! It's the W&B identifier, which is a `entity/project/name:version` string.

tests/utils.py
sfriedowitz: Ill rename this file in the follow up/

PR #92, opened 2024-04-04
Centralize artifact lineage creation on LMBuddy class

sfriedowitz:
## What's Changing

The key change here is that instead of logging artifact lineages as part of the job entrypoints, I am returning the raw W&B artifacts from the jobs, and logging them at the end of the runs in one location. This offers a few benefits:
1) We reduce a ton of boilerplate code duplication around when/where to initialize the W&B run. 
2) Because we are not logging artifacts within the job internals, we no longer need to inject `ArtifactLoader` interface into the entrypoints. This simplifies the code a good amount.
3) It further standardizes the input/output flow from the entrypoint functions, making them easier for developers to understand and write.
4) It begins to standardize the role of the `LMBuddy` class. In a follow-up PR, I plan to refactor the organization of the code a bit to add further functionality to this central class. Much of this is motivated by the design of libraries like Ludwig, and the way in which they use their `LudwigModel` central class.


examples/configs/evaluation/lm_harness_hf_config.yaml
```diff
@@ -1,3 +1,5 @@
+name: "lm-buddy-lm-harness"
```
sfriedowitz, line 1: I've decided to add a `name` field as a top-level parameter on on all `LMBuddyJobConfig`s. This is so we always have a consistent name for referring to the job run and any assets produced by it (e.g., the artifact names).

This is a pretty common pattern, if you take a look at MosaicML and how they handle their `RunConfig` object.

src/lm_buddy/integrations/wandb/run_config.py
```diff
@@ -21,12 +21,11 @@ class WandbRunConfig(BaseLMBuddyConfig):
     if it is not provided.
     """
 
-    __match_args__ = ("run_id", "name", "project", "run_group", "entity")
+    __match_args__ = ("id", "project", "group", "entity")
```
sfriedowitz, line 24: I've renamed these fields to exactly match the names of the fields in `wandb.init`. This is so we can dump a dict of this dataclass directly as `**kwargs` when calling that.

src/lm_buddy/jobs/_entrypoints/finetuning.py
```diff
@@ -67,35 +63,28 @@ def load_and_train(config: FinetuningJobConfig, artifact_loader: ArtifactLoader)
...
-        if is_tracking_enabled(config):
-            with wandb_init_from_config(
-                config.tracking,
-                resume=WandbResumeMode.NEVER,
-                job_type=LMBuddyJobType.FINETUNING,
-            ):
-                load_and_train(config, artifact_loader)
-        else:
-            load_and_train(config, artifact_loader)
+def training_function(config_data: dict[str, Any]):
```
sfriedowitz, line 66: Because we are no longer injecting the `ArtifactLoader`, this class doesn't need to be an internal method anymore. Minimal changes in the functionality here except for moving this function outside of the other method.

src/lm_buddy/jobs/configs/base.py
```diff
@@ -16,6 +20,15 @@ class LMBuddyJobConfig(BaseLMBuddyConfig):
     but this is not rigidly constrained by the interface. This may change in the future.
     """
 
+    name: str = Field(description="Name of the job.")
```
sfriedowitz, line 23: Making both of these inherited fields on `LMBuddyJobConfig`, since every job config should have them.

src/lm_buddy/jobs/configs/base.py
```diff
@@ -38,3 +51,12 @@ def to_tempfile(self, *, name: str = "config.yaml", dir: str | Path | None = Non
             config_path = Path(tmpdir) / name
             self.to_yaml_file(config_path)
             yield config_path
+
+    @abstractmethod
```
sfriedowitz, line 55: I was a bit unhappy with this method, but essentially I need a way of extracting all the _input_ `AssetPath`s off a `LMBuddyJobConfig`. This is so we can link them as input artifacts in the lineage generation.

The fields are not standard across jobs, so I left it abstract and forced every job config to implement it individually.

src/lm_buddy/buddy.py
```diff
@@ -16,25 +20,52 @@ class LMBuddy:
     Simple wrapper around executable functions for tasks available in the library.
     """
 
-    def __init__(self, artifact_loader: ArtifactLoader = WandbArtifactLoader()):
-        self._artifact_loader = artifact_loader
+    # TODO: Store some configuration (e.g., tracking info, name) globally on the buddy
+    def __init__(self):
+        pass
+
+    def _generate_artifact_lineage(
```
sfriedowitz, line 27: This is the central method that logs inputs to a run and logs output artifacts produced by a run.

imtihan (approved):

sfriedowitz:
Confirmed lineage generation working as expected on Ray: 
Screenshot 2024-04-04 at 9 56 33 AM

PR #93, opened 2024-04-04
Refactor module layout to separate configs and functionality

sfriedowitz:
## What's changing

Really, this PR is just moving a bunch of files around and fixing imports afterwards. The `integrations/wandb` and `integrations/huggingface` layout became tedious to work with IMO, so I think this is a more sensible approach that keeps things organized based on how/where they are used.

- Not much has changed in terms of functionality. **The CLI and the job functionality is 100% unchanged.**
- Refactor the internal module layout to something that I think makes a bit more sense. All configs are placed in the `configs/` module, and the associated helper methods and other functionality (e.g., the asset loader class) are moved closer to where they are used in the `jobs` folder.


imtihan (approved):
Generally looks good, the new code organization is more clear!

PR #94, opened 2024-04-04
Update dataset generation and storage for ragas and prometheus

sfriedowitz:
## What's changing

- Unifies how datasets are being stored between ragas and prometheus
- Imports `loguru` for logging. We just need to be careful not to accidentally send the logger across Ray remote boundaries, as that would incur serialization issues
- Adds a default storage path for lm-buddy. We can discuss this, and get rid of it if we don't feel its right.


src/lm_buddy/constants.py
```diff
@@ -0,0 +1,9 @@
+import os
+from pathlib import Path
+
+STORAGE_PATH_ENVIRONMENT_VARIABLE: str = "LM_BUDDY_STORAGE"
```
sfriedowitz, line 4: I kind of didn't like using the `RAY_STORAGE` value for the outputs generated by this library, since it's not even guaranteed to be running on Ray. 

Are we happy introducing our own `LM_BUDDY_STORAGE` env var for global control over outputs produced by our functions?

src/lm_buddy/jobs/evaluation/prometheus.py
```diff
@@ -107,25 +108,23 @@ def run_eval(config: PrometheusJobConfig) -> Path:
...
 
-    # enable / disable tqdm
+    # Enable / disable tqdm
     dataset_iterable = tqdm(dataset) if config.evaluation.enable_tqdm else dataset
 
-    # open the output file for writing and iterate on samples
-    tracking_name = config.tracking.name if config.tracking is not None else "output.json"
-    output_fname = Path(config.evaluation.output_folder) / tracking_name
-    with output_fname.open("w") as file:
+    # Generator that iterates over samples and yields new rows with the prometheus outputs
```
sfriedowitz, line 126: @aittalam I've opted to implement the `Dataset.from_generator` trick here to avoid having to write JSON to disk before generating the final dataset.

src/lm_buddy/jobs/evaluation/prometheus.py
```diff
@@ -107,25 +108,23 @@ def run_eval(config: PrometheusJobConfig) -> Path:
...
+    # Get the tokenizer
     tokenizer_config = AutoTokenizerConfig(path=config.prometheus.inference.engine)
     tokenizer = hf_loader.load_pretrained_tokenizer(tokenizer_config)
 
-    # enable / disable tqdm
+    # Enable / disable tqdm
     dataset_iterable = tqdm(dataset) if config.evaluation.enable_tqdm else dataset
 
-    # open the output file for writing and iterate on samples
-    tracking_name = config.tracking.name if config.tracking is not None else "output.json"
```
sfriedowitz, line 126: The `config.tracking.name` was a bug left in there by my previous refactors, that field no longer exists. Instead using `config.name` below to generate the directory path.

src/lm_buddy/jobs/evaluation/prometheus.py
```diff
@@ -107,25 +114,27 @@ def run_eval(config: PrometheusJobConfig) -> Path:
     # Instantiate OpenAI client to speak with the vLLM endpoint
     client = OpenAI(base_url=config.prometheus.inference.base_url)
 
-    # load dataset from W&B artifact
     hf_loader = HuggingFaceAssetLoader()
+
+    # Resolve the engine model
+    engine_path = hf_loader.resolve_asset_path(config.prometheus.inference.engine)
```
sfriedowitz, line 120: This was an issue left over after the path changes.

if the `engine` field has the path prefixes like `hf://` or `wandb://`, then it has to be resolved before passing to the OpenAI client for completions.

src/lm_buddy/constants.py
```diff
@@ -0,0 +1,7 @@
+import os
+from pathlib import Path
+
+DEFAULT_STORAGE_PATH: str = os.getenv(
```
sfriedowitz, line 4: Are we happy introducing our own `LM_BUDDY_STORAGE` env var for global control over outputs produced by our functions? It didn't feel right using the `RAY_STORAGE` path, since this code may not even be running on Ray.

sfriedowitz: Maybe I'll just leave the defaults to `/tmp` for the `storage_path` on the configs. That would be fine for now too.

imtihan: I think for now let's leave it. We might eventually have staging and deployed environments, we can refactor this at that time.

imtihan (approved):
LGTM!

src/lm_buddy/jobs/evaluation/prometheus.py
aittalam: Thanks! It looks like `Dataset.from_generator` [saves the output incrementally](https://huggingface.co/docs/datasets/create_dataset#from-local-files) and then performs memory-mapping. This sounds good as what I was afraid of was what happened if the process broke right before the end (these evals can take hours even on GPUs). As mostly a note to self, we should simulate an issue and see if/how data can be recovered from an interrupted run.

src/lm_buddy/jobs/evaluation/prometheus.py
```diff
@@ -84,18 +88,21 @@ def instruction_to_prompt(config: PrometheusJobConfig, instruction: str) -> str:
 
 
 def get_response_with_retries(
-    config: PrometheusJobConfig, client: OpenAI, prompt: str, max_retries: int
+    config: PrometheusJobConfig,
+    client: OpenAI,
+    engine: str,
+    prompt: str,
```
aittalam, line 94: Thanks for catching the redundant max_retries! It was a byproduct of tests I did before putting it in the config and then I missed it when refactoring 🙏

aittalam (approved):
Thanks for working on this, Sean!
I think both the common storage path and the new dataset generation are nice features to have.
My only concern is what happens when the process breaks and if/how data can be recovered, but I saw Dataset.from_generator still saves incrementally by default and we can think about recovery after doing some tests with it

src/lm_buddy/jobs/evaluation/prometheus.py
sfriedowitz: Also, we can likely use the batched map and parallelization tools from Datasets to speed up iteration over large datasets.

sfriedowitz: Let's come back to this in a follow-up if we so desire. Easy fixes using `from_generator`, either way.

PR #99, opened 2024-05-17
Added S3 support for datasets, refactored asset loaders

aittalam:
## What's changing

- Added S3 support for datasets
- Refactored the asset loaders to (1) support S3 datasets and (2) be asset-specific. We now have one class for datasets, one for models/configs, one for tokenizers.

## How to test it
- Used pytest for backwards compatibility.
- Manually tested S3 dataset loading in the prometheus job (we might want to rely on localstack for testing)

### DIFF
--- a/src/lm_buddy/jobs/asset_loader.py
+++ b/src/lm_buddy/jobs/asset_loader.py
@@ -27,8 +27,10 @@ from lm_buddy.tracking.artifact_utils import get_artifact_directory, get_artifac
 class HuggingFaceAssetLoader:
     """Helper class for loading HuggingFace assets from LM Buddy configurations.
 
-    TODO: We can probably move these to standalone functions now that ArtifactLoader is gone.
-          What if we add other deps (e.g, S3 client in the future?)
+    The base class provides implementation for resolving the asset path.
+    Resolution depends both on the source and on the artifact type.
+    TODO: implement resolution for non-dataset artifacts
+
     """
 
     def resolve_asset_path(self, path: AssetPath) -> str:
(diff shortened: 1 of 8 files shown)


src/lm_buddy/jobs/asset_loader.py
```diff
@@ -27,8 +27,10 @@
 class HuggingFaceAssetLoader:
```
sfriedowitz, line 27: After I removed the "artifact loader" interface a few PRs ago, these classes no longer carry state (per the deleted comment below). It might be more fitting to simply turn these into functions instead of classes.

The reason I could see a class being useful is if the asset loader had some external dependency (e.g. an S3 client from boto3) that is required to do its job. But right now, thats not the case.

src/lm_buddy/jobs/evaluation/lm_harness.py
```diff
@@ -37,10 +37,10 @@ def get_per_task_dataframes(
 
 def load_harness_model(config: LMHarnessJobConfig) -> HFLM | OpenaiCompletionsLM:
     # Instantiate the lm-harness LM class based on the provided model config type
-    hf_loader = HuggingFaceAssetLoader()
+    hf_model_loader = HuggingFaceModelLoader()
```
sfriedowitz, line 40: Same thing w/ my comment above. These classes have no internal state, so a function seems a bit cleaner.

sfriedowitz (commented):
I have one comment to consider regarding asset loader classes vs helper functions. It's your call, so if you don't want to make the changes I can approve after your response.

aittalam:
> I have one comment to consider regarding asset loader classes vs helper functions. It's your call, so if you don't want to make the changes I can approve after your response.

Thanks Sean! As S3 support for models & co (basically anything which is not a dataset) has a dependency on an s3 client, I think it makes sense to keep this (even if I agree with you that the function approach looks cleaner!).

sfriedowitz (approved):
LGTM

sfriedowitz:
@aittalam whoops! Add a version bump before merging please.

aittalam:
> @aittalam whoops! Add a version bump before merging please.

Great catch! 🙏 :-)

PR #100, opened 2024-06-12
New huggingface eval for the summarization use case with rouge, meteor, and bertscore

aittalam:
## What's changing
Added a new `evaluate huggingface` entrypoint which supports evaluation of local and remote models (seq2seq, causal, openai, vllm, llamafile) and loading datasets / saving results on s3.

## How to test it
```
lm-buddy evaluate huggingface --config examples/configs/evaluation/hf_evaluate_config.yaml
```

## Related Jira Ticket
https://mzai.atlassian.net/browse/MZPLATFORM-78

I know we discussed messaging to mzai-platform's backend directly from lm-buddy jobs.
I am 100% in favor of it, I just wanted to keep this PR independent from messaging and I will create a new one to be tested together with the updated mzai-platform code.

### DIFF
--- a/examples/configs/evaluation/hf_evaluate_config.yaml
+++ b/examples/configs/evaluation/hf_evaluate_config.yaml
@@ -1,25 +1,65 @@
 name: "lm-buddy-hf-evaluate"
 
+# Input dataset path
 dataset:
   path: "s3://platform-storage/datasets/dialogsum"
 
-# Settings specific to hf_evaluate
-evaluation:
-  metrics: ["rouge", "bertscore", "meteor"]
-  # enable/disable tqdm to track eval progress
-  enable_tqdm: False
-  # rely on HF pipeline for summarization
-  use_pipeline: True
 
-# Model to evaluate
+# Settings specific to the hf_evaluate entrypoint
+evaluation:
+  # metrics to be used for the evaluation
+  # (you can add "rouge", "meteor", and "bertscore" atm)
+  metrics: ["rouge", "meteor"]
+  # enable/disable tqdm to track eval progress
+  # (useful when running interactively, noisy on ray logs)
+  enable_tqdm: True
+  # rely on HF pipeline for summarization (ignored if using OAI API)
+  use_pipeline: True
+  # perform inference / evaluation on the first max_samples only
+  max_samples: 10
+  # output file path
+  # - if you provide a path complete with a filename, results will be stored in it
+  # - if you provide a dir, results will be stored in <dir>/<config.name>/eval_results.json
+  # - if you don't provide a storage path, results will be stored locally (see ~/.lm-buddy/results)
+  # storage_path: "s3://platform-storage/experiments/results/"
+
+# Model to evaluate. Choose one of the following options by uncommenting it
+
+# 1. Local model
+#    - Provide model path to load the model locally
+#    - Make sure you add quantization details (see below) if the model is too large
+#    - Optionally, add a tokenizer (the one matching the specified model name is the default)
 model:
   path: "hf://facebook/bart-large-cnn"
 
-quantization:
-  load_in_4bit: True
-  bnb_4bit_quant_type: "fp4"
+# # 2. OpenAI
+# #    - The base_url is fixed
+# #    - Choose an engine name (see https://platform.openai.com/docs/models)
+# #    - Customize the system prompt if needed
+# model:
+#   inference:
+#     base_url: "https://api.openai.com/v1"
+#     engine: "oai://gpt-4-turbo"
+#     system_prompt: "You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences."
+#     max_retries: 3
 
-# # Tracking info for where to log the run results
+# # 3. OpenAI - compatible model
+# #    - Works with local/remote vLLM-served models and llamafiles
+# #    - Provide base_url and engine
+# #    - Customize the system prompt if needed
+# model:
+#   inference:
+#     base_url: "http://localhost:8081/v1"
+#     engine: "hf://mistralai/mistral-7b-instruct-v0.2"
+#     system_prompt: "You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences."
+#     max_retries: 3
+
+# Quantization (use it if you are dealing with models too large to fit in RAM)
+# quantization:
+#   load_in_4bit: True
+#   bnb_4bit_quant_type: "fp4"
+
+# Tracking info for where to log the run results
 # tracking:
 #   project: "lm-buddy-examples"
 #   entity: "sample"
--- a/examples/configs/evaluation/hf_evaluate_config.yaml
+++ b/examples/configs/evaluation/hf_evaluate_config.yaml
@@ -1,0 +1,25 @@
+name: "lm-buddy-hf-evaluate"
+
+dataset:
+  path: "s3://platform-storage/datasets/dialogsum"
+
+# Settings specific to hf_evaluate
+evaluation:
+  metrics: ["rouge"]
+  # enable/disable tqdm to track eval progress
+  enable_tqdm: False
+  # rely on HF pipeline for summarization
+  use_pipeline: True
+
+# Model to evaluate
+model:
+  path: "hf://facebook/bart-large-cnn"
+
+quantization:
+  load_in_4bit: True
+  bnb_4bit_quant_type: "fp4"
+
+# # Tracking info for where to log the run results
+# tracking:
+#   project: "lm-buddy-examples"
+#   entity: "sample"
--- a/examples/configs/evaluation/hf_evaluate_config.yaml
+++ b/examples/configs/evaluation/hf_evaluate_config.yaml
@@ -5,7 +5,7 @@ dataset:
 
 # Settings specific to hf_evaluate
 evaluation:
-  metrics: ["rouge"]
+  metrics: ["rouge", "bertscore", "meteor"]
   # enable/disable tqdm to track eval progress
   enable_tqdm: False
   # rely on HF pipeline for summarization
--- a/src/lm_buddy/buddy.py
+++ b/src/lm_buddy/buddy.py
@@ -3,12 +3,14 @@ import wandb
 from lm_buddy.configs.jobs import (
     EvaluationJobConfig,
     FinetuningJobConfig,
+    HuggingFaceEvalJobConfig,
     JobConfig,
     LMHarnessJobConfig,
     PrometheusJobConfig,
     RagasJobConfig,
 )
 from lm_buddy.jobs.common import EvaluationResult, FinetuningResult, JobType
+from lm_buddy.jobs.evaluation.hf_evaluate import run_hf_evaluation
 from lm_buddy.jobs.evaluation.lm_harness import run_lm_harness
 from lm_buddy.jobs.evaluation.prometheus import run_prometheus
 from lm_buddy.jobs.evaluation.ragas import run_ragas
@@ -66,6 +68,8 @@ class LMBuddy:
                 result = run_prometheus(prometheus_config)
             case RagasJobConfig() as ragas_config:
                 result = run_ragas(ragas_config)
+            case HuggingFaceEvalJobConfig() as hf_eval_config:
+                result = run_hf_evaluation(hf_eval_config)
             case _:
                 raise ValueError(f"Invlid configuration for evaluation: {type(config)}")
         self._generate_artifact_lineage(config, result.artifacts, JobType.EVALUATION)
--- a/src/lm_buddy/configs/jobs/hf_evaluate.py
+++ b/src/lm_buddy/configs/jobs/hf_evaluate.py
@@ -1,0 +1,70 @@
+from pydantic import Field, conlist, field_validator, model_validator
+
+from lm_buddy.configs.common import LMBuddyConfig
+from lm_buddy.configs.huggingface import (
+    AutoModelConfig,
+    AutoTokenizerConfig,
+    DatasetConfig,
+    QuantizationConfig,
+)
+from lm_buddy.configs.jobs.common import JobConfig
+from lm_buddy.configs.vllm import VLLMCompletionsConfig
+from lm_buddy.paths import AssetPath
+
+
+class HuggingFaceEvaluationConfig(LMBuddyConfig):
+    """Misc settings provided to an lm-harness evaluation job."""
+
+    metrics: conlist(str, min_length=1)
+    use_pipeline: bool = False
+    enable_tqdm: bool = False
+
+
+class HuggingFaceEvalJobConfig(JobConfig):
+    """Configuration to run a HuggingFace evaluation job."""
+
+    dataset: DatasetConfig = Field(
+        description="Dataset of text completions to evaluate using the Prometheus judge model."
+    )
+    evaluation: HuggingFaceEvaluationConfig
+    model: AutoModelConfig | VLLMCompletionsConfig
+    quantization: QuantizationConfig | None = None
+    tokenizer: AutoTokenizerConfig
+
+    @model_validator(mode="before")
+    def ensure_tokenizer_config(cls, values):
+        """Set the tokenizer to the model path when not explicitly provided."""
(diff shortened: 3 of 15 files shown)


src/lm_buddy/configs/jobs/hf_evaluate.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+    evaluation: HuggingFaceEvaluationConfig
+    model: AutoModelConfig | VLLMCompletionsConfig
+    quantization: QuantizationConfig | None = None
+    tokenizer: AutoTokenizerConfig
+
+    @model_validator(mode="before")
+    def ensure_tokenizer_config(cls, values):
+        """Set the tokenizer to the model path when not explicitly provided."""
+        if values.get("tokenizer") is None:
```
veekaybee, line 39: We might have previous code for this that we can use here instead? https://github.com/mozilla-ai/lm-buddy/blob/72d617963f0f83c568d32615d59d3f30403d6b5f/src/lm_buddy/jobs/asset_loader.py#L135

src/lm_buddy/jobs/evaluation/hf_evaluate.py
```diff
@@ -0,0 +1,115 @@
...
+    t = time.time()
+    for sample_txt in dataset_iterable:
+        predictions.append(model_client.predict(sample_txt))
+    summarization_time = time.time() - t
+    logger.info(f"Summarization performed in {summarization_time} seconds")
+
+    # run evaluation
+    ground_truth = dataset["ground_truth"]
+    em = EvaluationMetrics(config.evaluation.metrics)
+    t = time.time()
```
veekaybee, line 98: we set t=time twice here,one might clobber the other in case both are run. consider putting at top of method call or reassigning variable name

examples/configs/evaluation/hf_evaluate_config.yaml
```diff
@@ -0,0 +1,65 @@
...
+# Input dataset path
+dataset:
+  path: "s3://platform-storage/datasets/dialogsum"
+
+
+# Settings specific to the hf_evaluate entrypoint
+evaluation:
+  # metrics to be used for the evaluation
+  # (you can add "rouge", "meteor", and "bertscore" atm)
+  metrics: ["rouge", "meteor"]
```
veekaybee, line 12: should we set this to the full list of metrics as a default?

examples/configs/evaluation/hf_evaluate_config.yaml
```diff
@@ -0,0 +1,65 @@
...
+# Model to evaluate. Choose one of the following options by uncommenting it
+
+# 1. Local model
+#    - Provide model path to load the model locally
+#    - Make sure you add quantization details (see below) if the model is too large
+#    - Optionally, add a tokenizer (the one matching the specified model name is the default)
+model:
+  path: "hf://facebook/bart-large-cnn"
+
+# # 2. OpenAI
```
veekaybee, line 35: let's have one level of comments for this file

src/lm_buddy/configs/jobs/hf_evaluate.py
```diff
@@ -0,0 +1,83 @@
...
+)
+from lm_buddy.configs.jobs.common import JobConfig
+from lm_buddy.configs.vllm import VLLMCompletionsConfig
+from lm_buddy.paths import AssetPath
+
+
+class HuggingFaceEvaluationConfig(LMBuddyConfig):
+    """Misc settings provided to an lm-harness evaluation job."""
+
+    metrics: conlist(str, min_length=1)
```
veekaybee, line 18: TIL conlist!

src/lm_buddy/configs/vllm.py
```diff
@@ -12,12 +12,18 @@ class InferenceServerConfig(LMBuddyConfig):
...
     """
 
     base_url: str
     engine: AssetPath
 
+    # optional system prompt to be used by default in chat completions
+    system_prompt: str | None = None
+
+    # max number of retries when communication with server fails
+    max_retries: int | None = None
```
veekaybee, line 25: should this be set to a default int?

src/lm_buddy/jobs/asset_loader.py
```diff
@@ -120,7 +125,21 @@ def load_pretrained_model(
...
+        # print(model_config)
+
+        if getattr(model_config, "model_type") in MODEL_FOR_SEQ_TO_SEQ_CAUSAL_LM_MAPPING_NAMES:
+            automodel_class = AutoModelForSeq2SeqLM
+        elif getattr(model_config, "model_type") in MODEL_FOR_CAUSAL_LM_MAPPING_NAMES:
+            automodel_class = AutoModelForCausalLM
+        else:
+            logger.info("Model type not supported. Trying AutoModelForCausalLM")
+            automodel_class = AutoModelForCausalLM
+        # print(automodel_class)
```
veekaybee, line 140: nit: remove

src/lm_buddy/jobs/evaluation/metrics.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # initialize dictionary with metric name
+        evals = {"meteor": []}
+
+        # run sample-wise evals (as default implementation only returns mean value)
+        for p, r in zip(pred, ref):
+            evals["meteor"].append(ev.compute(predictions=[p], references=[r])["meteor"])
+
+        # calculate mean
+        evals[f"meteor_mean"] = np.mean(evals["meteor"])
```
binaryaaron, line 48: +1 to the linter, just remove the `f`.
```suggestion
        evals["meteor_mean"] = np.mean(evals["meteor"])
```

src/lm_buddy/jobs/evaluation/hf_evaluate.py
```diff
@@ -0,0 +1,115 @@
...
+    evaluation_results = em.run_all(predictions, ground_truth)
+    summarization_time = time.time() - t
+    logger.info(f"Summarization performed in {summarization_time} seconds")
+
+    return save_outputs(config, evaluation_results)
```
veekaybee, line 103: in general we wouldn't expect a save_outputs function to return anything so it might make sense to rename or do something like 

output_paths = save_outputs(config, evaluation_results)
return output_paths

src/lm_buddy/jobs/evaluation/hf_evaluate.py
```diff
@@ -0,0 +1,115 @@
...
+        if storage_path.endswith("/"):
+            storage_path = "s3://" + str(Path(storage_path[5:]) / config.name / "eval_results.json")
+        logger.info(f"Storing into {storage_path}...")
+        s3.put_file(local_path, storage_path)
+        return storage_path
+    else:
+        return local_path
+
+
+def run_eval(config: HuggingFaceEvalJobConfig) -> Path:
```
veekaybee, line 51: should the return type be a list of Paths?

src/lm_buddy/jobs/model_clients.py
```diff
@@ -0,0 +1,111 @@
...
+            top_p=config.top_p,
+        )
+
+    def _get_response_with_retries(
+        self,
+        config: VLLMCompletionsConfig,
+        prompt: str,
+    ) -> tuple[str, str]:
+        current_retry_attempt = 1
+        max_retries = 1 if config.inference.max_retries is None else config.inference.max_retries
```
binaryaaron, line 94: @veekaybee there's a default for retries here

src/lm_buddy/jobs/evaluation/hf_evaluate.py
binaryaaron: yeah, could make a small timing decorator or context to use and log these more cohesively.

src/lm_buddy/buddy.py
```diff
@@ -3,12 +3,14 @@
...
     EvaluationJobConfig,
     FinetuningJobConfig,
+    HuggingFaceEvalJobConfig,
     JobConfig,
     LMHarnessJobConfig,
     PrometheusJobConfig,
     RagasJobConfig,
 )
 from lm_buddy.jobs.common import EvaluationResult, FinetuningResult, JobType
+from lm_buddy.jobs.evaluation.hf_evaluate import run_hf_evaluation
```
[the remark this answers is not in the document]
binaryaaron: it's a natural evolution of the api; i'd opt for that to be done in a followup unless it's pretty trivial. worth looking at though; might help with the eventual sdk/cli work to be done too.

src/lm_buddy/configs/vllm.py
binaryaaron: a default is set downstream; tagged you in a comment.

binaryaaron (commented):
mostly responding / adding on to @veekaybee's review; will look more later. thanks!

src/lm_buddy/jobs/evaluation/hf_evaluate.py
```diff
@@ -0,0 +1,115 @@
...
+from lm_buddy.jobs.evaluation.metrics import EvaluationMetrics
+from lm_buddy.jobs.model_clients import (
+    HuggingFaceModelClient,
+    OpenAIModelClient,
+    PipelineModelClient,
+)
+
+
+def save_outputs(config: HuggingFaceEvalJobConfig, evaluation_results: dict) -> Path:
+    storage_path = config.evaluation.storage_path
```
veekaybee, line 30: We might have two different functions here to save locally and save to the cloud and abstract away into a try/except, i.e. 

```
def save_to_disk():
      ```save local output```

def save_to_s3():
      ```save to s3 output```
      
try:
   save_local()
   save_cloud()
except:
   logger.info("Wasn't able to save")
 ```

File: src/lm_buddy/jobs/evaluation/metrics.py [hunk 14]
```diff
@@ -0,0 +1,70 @@
...
+
+class EvaluationMetrics:
+    def __init__(self, metrics):
+        self._supported_metrics = {
+            "rouge": self._rouge,
+            "meteor": self._meteor,
+            "bertscore": self._bertscore,
+        }
+
+        self._chosen_metrics = set(metrics).intersection(set(self._supported_metrics.keys()))
```
veekaybee, line 14: we could make this a bit easier to read and change it to :

```set(metrics) & supported_metrics.keys()```

File: examples/configs/evaluation/hf_evaluate_config.yaml [hunk 15]
```diff
@@ -0,0 +1,65 @@
...
+  use_pipeline: True
+  # perform inference / evaluation on the first max_samples only
+  max_samples: 10
+  # output file path
+  # - if you provide a path complete with a filename, results will be stored in it
+  # - if you provide a dir, results will be stored in <dir>/<config.name>/eval_results.json
+  # - if you don't provide a storage path, results will be stored locally (see ~/.lm-buddy/results)
+  # storage_path: "s3://platform-storage/experiments/results/"
+
+# Model to evaluate. Choose one of the following options by uncommenting it
```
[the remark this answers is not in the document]
aittalam: You mean switch between differente models? I mainly wanted to provide a configuration showcasing all the possibilities, because we have been supporting different models (in other entrypoints too) but a single example is not very self-explicative.
I think cli params are a potential alternative, wdyt would work best if we think about running these from mzai-platform?

src/lm_buddy/buddy.py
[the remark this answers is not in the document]
aittalam: Yeah, I think the more we add the more we'll want to factor common code together - and this involves metrics we might want to share, how we run model inference (right now we do it in 3 different ways across lm-h, prometheus, and this), common HF functionalities, etc.
I am 100% for doing it in a follow-up :-) I hope I am not overcomplicating it, but I think if we shape it properly it'll be a good investment

src/lm_buddy/configs/jobs/hf_evaluate.py
aittalam: Whoops, not sure I understood your comment: we do call `load_pretrained_tokenizer` in the entrypoint code, `ensure_tokenizer_config` just makes sure there is always an `AutoTokenizerConfig` to pass even when there is none provided (in that case it will just use the model path as the tokenizer path).
WDYT about the choice of constraining the config class to `AutoTokenizerConfig`? I appreciate the strict typing even if it requires some extra validation as the above, but what I am not 100% happy with is that, depending on the application, we might have some cases where we do not need a tokenizer at all (e.g. when using an openai-compatible API) and we still have to go through generating one.

src/lm_buddy/jobs/evaluation/hf_evaluate.py
aittalam: Good catch, TY!

examples/configs/evaluation/hf_evaluate_config.yaml
aittalam: Yeah, my bad, bertscore is slower and I left it out while heavily testing, then forgot to put it back before committing 😅 - I'll bring it back!

examples/configs/evaluation/hf_evaluate_config.yaml
aittalam: Ok! We can also make more than one config file for the same entrypoint to show the different options, so nothing will have to be commented / decommented

src/lm_buddy/jobs/evaluation/hf_evaluate.py
aittalam: My thought was one model + one evaluation (w multiple metrics) => one output path.

Were you referring to the possibility of evaluating multiple models? My rationale for keeping it atomic is that we'll likely want to parallelize evals on different models with different ray jobs, thus the logic for splitting the higher-level task in multiple evals (and also for retrieving different eval results and presenting them together) should be in the mzai-platform backend. WDYT?

src/lm_buddy/jobs/evaluation/metrics.py
aittalam: oof forgot it after changing from parametric code iterating on all metrics to manual one (meteor has just one metric. Good catch, thanks! (and thanks linter ;-))

src/lm_buddy/jobs/model_clients.py
aittalam: Btw I am happy to put the default in config too, perhaps it is cleaner? I would leave it in the code in any case, as you always have at least 1 attempt.

src/lm_buddy/jobs/evaluation/hf_evaluate.py
aittalam: Yeah that would be cleaner - I already miss our artifacts so much! Are we sure we don't want mlflow to manage all of this? 😅

examples/configs/evaluation/hf_evaluate_config.yaml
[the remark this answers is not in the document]
aittalam: Will address this in one of the follow-up comments!

src/lm_buddy/jobs/evaluation/hf_evaluate.py
aittalam: Done, added a decorator which logs + returns timing so we can save it in the output file too (we'll likely want to use this information at some point)

src/lm_buddy/configs/jobs/hf_evaluate.py
veekaybee: Yep, that makes sense - my comment was more about the fact that we do this check several times in the platform IIRC and might make sense to have a utility function, but can be addressed in a follow-up

aittalam: Great, thanks! 🙏

src/lm_buddy/jobs/evaluation/hf_evaluate.py
veekaybee: ah yep, my bad! here mostly I was referring to this return `return save_outputs(config, evaluation_results)` which I thought returned both paths.

aittalam: Ah that makes sense! Yeah, the local one for me was mostly a temp file for the s3 one. It definitely comes useful for local testing, but will eventually be lost when ray jobs run lm-buddy

veekaybee (approved):
Thanks for addressing these and for this PR!

PR #103, opened 2024-06-16
Add input data + predictions to output file

aittalam:
## What's changing

Added the possibility of storing both input data and predictions to the output json file

### DIFF
--- a/examples/configs/evaluation/hf_evaluate_config.yaml
+++ b/examples/configs/evaluation/hf_evaluate_config.yaml
@@ -21,6 +21,10 @@ evaluation:
   # - if you provide a dir, results will be stored in <dir>/<config.name>/eval_results.json
   # - if you don't provide a storage path, results will be stored locally (see ~/.lm-buddy/results)
   # storage_path: "s3://platform-storage/experiments/results/"
+  # return input data in the output file
+  return_input_data: True
+  # return predictions in the output file
+  return_predictions: True
 
 # Model to evaluate (local).
 # - Provide model path to load the model locally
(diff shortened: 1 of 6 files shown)


veekaybee (approved):
The code looks good but I was curious about the outputs + also looks like tests need a bump

aittalam:
> How big do these end up being and what does the final output file look like in this case? it would be nice to have some output examples!

Great point! Here's an output example.
If one adds everything the file might become quite large: this one is 12KB for 10 samples, the dialogsum validation which has 500 samples should become about 600KB. This is the reason why I wanted to add the option to only have the eval metrics saved. In any case, if the input size is D, the output dataset with everything will be of size E = D + e where e is typically quite smaller than D as the new data are metrics and summarized texts.


[eval_results.json](https://github.com/user-attachments/files/15888984/eval_results.json)

aittalam:
> The code looks good but I was curious about the outputs + also looks like tests need a bump

For the integration tests: I saw the error 

```
"cannot import name 'packaging' from 'pkg_resources'"
```

Occurs a few times and every time it starts from some ray code. So I thought perhaps the problem was not in setuptools as I originally thought (see the plethora of posts [here](https://duckduckgo.com/?t=ffab&q=%22cannot+import+name+%27packaging%27+from+%27pkg_resources%27%22&ia=web)) but rather in ray. 
I looked into one of the failing scripts ([dashboard/modules/dashboard_sdk.py](https://github.com/ray-project/ray/commit/7e9043c38d76412c310fcf6e3fff79cb55d481da)) and found that the latest update to it was exactly removing the [deprecated pkg_resources](https://github.com/pypa/setuptools/blob/22993191cfea75c7f53ca2d27962c014511f0d92/pkg_resources/__init__.py#L19) in favor of importlib.metadata.

This happened in ray version 2.24.0 and currently we have 2.9.3 set in pyproject.toml. @chainlink is there anything against us spinning up ray clusters with a more recent ray version, so we can update it here too?

PR #105, opened 2024-06-25
Fix index out of bound error if input size is greater than the maximu…

imtihan:
…m context length 

## What's changing
parameter value on tokenization

## How to test it
run an lm-buddy hf-evaluate job with the `pegasus-xsum` model and email thread summary (https://huggingface.co/datasets/sidhq/email-thread-summary) dataset. It should run without errors.


binaryaaron (commented):
is this in draft still?

aittalam:
Addressed in [PR113](https://github.com/mozilla-ai/lm-buddy/pull/113)

PR #107, opened 2024-07-01
Bumped python, pydantic, version 0.10.9

aittalam:
## What's changing

- Update to python 3.11
- Updated pydantic to 2.6.4
- Bumped version

### DIFF
--- a/src/lm_buddy/paths.py
+++ b/src/lm_buddy/paths.py
@@ -68,16 +68,16 @@ AssetPath = Annotated[str, AfterValidator(lambda x: validate_asset_path(x))]
 
 def format_file_path(path: str | Path) -> AssetPath:
     path = Path(path).absolute()
-    return f"{PathPrefix.FILE}{path}"
+    return f"{PathPrefix.FILE.value}{path}"
 
 
 def format_huggingface_path(repo_id: str) -> AssetPath:
-    return f"{PathPrefix.HUGGINGFACE}{repo_id}"
+    return f"{PathPrefix.HUGGINGFACE.value}{repo_id}"
 
 
 def format_artifact_path(artifact: wandb.Artifact) -> AssetPath:
     try:
-        return f"{PathPrefix.WANDB}{artifact.qualified_name}"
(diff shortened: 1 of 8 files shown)


aittalam:
I ran unit / integration tests:

- the good news is that with the python bump the lm-harness integration error disappeared (I guess Vicki was already running 3.11)
- the bad bews is that there was some change in how class(str, Enum) are treated in f strings, so if you `print(f"{PathPrefix.FILE}")` you get `PathPrefix.FILE` and to get `file://` you need to `print(f"{PathPrefix.FILE.value}")`. I fixed it and now all unit tests and integration tests run properly
- TL/DR is: things should be fine now, but we need more unit/integration tests (especially for the latest eval code)

aittalam:
I have updated github workflows and tests should be successful now (oof I take it back - lm-harness breaks again even if now on my laptop it runs properly).

There still remains an open point, i.e. the docs refer to the current Ray cluster running python 3.10.* and we should update this when we have updated our ray cluster too. Will get back on it as soon as we know the exact python version we run on the ray cluster

aittalam:
Updated python version on ray in source (3.11.9) and documentation here. It should all be consistent now 🤞

src/lm_buddy/paths.py
```diff
@@ -68,16 +68,16 @@ def validate_asset_path(path: str) -> "AssetPath":
 
 def format_file_path(path: str | Path) -> AssetPath:
     path = Path(path).absolute()
-    return f"{PathPrefix.FILE}{path}"
+    return f"{PathPrefix.FILE.value}{path}"
```
davidmanzanoai, line 71: We need to have a unit test to validate these kind of functions. Do we have any?

aittalam: Yes! I think they were essential to catch these issues, and I'll make sure to create more going on

veekaybee (approved):
Chatted about this, LGTM minus any  potential issues caused by transformers bump, leaving as a future note

PR #108, opened 2024-07-02
Updated prometheus job to get rid of fschat dependency

aittalam:
## What's changing

Removed the dependency on fastchat, which fixes a lockfile generation down the line due to a chain of imports ending in a library (wavedrom) with bad metadata / setup file.

For now I just imported one single script from an older version of the library, then we'll get rid of it for good as prometheus only needs llama-2 conversation template

## How to test it

- `lm-buddy evaluate prometheus --config <a valid config file>` (more details in https://asciinema.org/a/FiYe21Ld7vRv7ouwcVtIRIdB8) - tested locally and working
- `pytest`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "lm-buddy"
-version = "0.10.9"
+version = "0.10.10"
 authors = [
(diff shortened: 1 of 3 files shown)


binaryaaron (approved):
@aittalam  and i talked about this one directly. thanks; 🚢 .

I'd add a noqa to the whole `conversations.py` file and call it good.

PR #109, opened 2024-07-03
Makes deps that require torch optional by default

binaryaaron:
## What's changing
As discussed, this changes the dependency structure of this repo to make all packages that require Torch optional by default. This allows reuse of the configs more readily without other structural changes.

## How to test it
Tested internally by generating lockfiles on a project requiring this project with different  combos of [torchy, test, docs], verified pytorch is only in the ones in `torchy`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "lm-buddy"
-version = "0.10.8"
+version = "0.11.0"
 authors = [
     { name = "Sean Friedowitz", email = "sean@mozilla.ai" },
     { name = "Aaron Gonzales", email = "aaron@mozilla.ai" },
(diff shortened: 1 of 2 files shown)


binaryaaron:
As a note, we can fix the flakey integration test - will do in a distinct PR.

pyproject.toml
```diff
@@ -20,18 +20,25 @@ requires-python = ">=3.11,<3.12"
...
     "pydantic-yaml==1.2.0",
     "ray[default]==2.30.0",
     "loguru==0.7.2",
     "s3fs",
-    # HuggingFace
+]
+
+[project.optional-dependencies]
+ruff = ["ruff==0.2.1"]
+torchy = [
```
veekaybee, line 39: maybe `jobs`  since it's `evaluation` + `finetuning`?

veekaybee (approved):
👍 lgtm

pyproject.toml
binaryaaron: i suppose that is indeed better than `torchy`. :)

PR #112, opened 2024-07-30
Mixed fixes + support for Mistral API

aittalam:
Three main commits:
- added `allow_remote_code` to config => allows us to run Phi models
- fixed `max_samples` bug (now if the value is 0 all samples are considered, if it is larger than len(dataset) we get a warning and all the dataset is used)
- added support for Mistral API (created a general `APIModelClient` class and made OAI and Mistral as two subclasses as so much code was in common)
- bonus: added model name in output evals file :-)

Tested manually with different configs (seq2seq models locally, OAI, Mistral) and with pytest.

### DIFF
--- a/src/lm_buddy/jobs/asset_loader.py
+++ b/src/lm_buddy/jobs/asset_loader.py
@@ -45,7 +45,9 @@ class HuggingFaceAssetLoader:
         The returned string has its `PathPrefix` stripped away..
         """
         raw_path = strip_path_prefix(path)
-        if path.startswith((PathPrefix.FILE, PathPrefix.HUGGINGFACE, PathPrefix.OPENAI)):
+        if path.startswith(
+            (PathPrefix.FILE, PathPrefix.HUGGINGFACE, PathPrefix.OPENAI, PathPrefix.MISTRAL)
+        ):
             return raw_path
         elif path.startswith(PathPrefix.WANDB):
             artifact = get_artifact_from_api(raw_path)
--- a/src/lm_buddy/jobs/asset_loader.py
+++ b/src/lm_buddy/jobs/asset_loader.py
@@ -97,7 +97,9 @@ class HuggingFaceModelLoader(HuggingFaceAssetLoader):
         An exception is raised if the HuggingFace repo does not contain a `config.json` file.
         """
         config_path = self.resolve_asset_path(config.path)
-        return AutoConfig.from_pretrained(pretrained_model_name_or_path=config_path)
+        return AutoConfig.from_pretrained(
+            pretrained_model_name_or_path=config_path, trust_remote_code=config.trust_remote_code
+        )
 
     def load_pretrained_model(
         self,
--- a/src/lm_buddy/jobs/evaluation/hf_evaluate.py
+++ b/src/lm_buddy/jobs/evaluation/hf_evaluate.py
@@ -91,7 +91,7 @@ def run_eval(config: HuggingFaceEvalJobConfig) -> Path:
 
     # Limit dataset length if max_samples is specified
     max_samples = config.evaluation.max_samples
-    if max_samples is not None and max_samples > 0:
+    if max_samples and max_samples > 0:
         if max_samples > len(dataset):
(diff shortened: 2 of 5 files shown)


src/lm_buddy/jobs/asset_loader.py
```diff
@@ -97,7 +99,9 @@ def load_pretrained_config(
         An exception is raised if the HuggingFace repo does not contain a `config.json` file.
         """
         config_path = self.resolve_asset_path(config.path)
-        return AutoConfig.from_pretrained(pretrained_model_name_or_path=config_path)
+        return AutoConfig.from_pretrained(
+            pretrained_model_name_or_path=config_path, trust_remote_code=config.trust_remote_code
```
veekaybee, line 103: ❤️

src/lm_buddy/jobs/evaluation/hf_evaluate.py
```diff
@@ -89,8 +90,12 @@ def run_eval(config: HuggingFaceEvalJobConfig) -> Path:
     dataset = hf_dataset_loader.load_dataset(config.dataset)
 
     # Limit dataset length if max_samples is specified
-    if config.evaluation.max_samples is not None:
-        dataset = dataset.select(range(config.evaluation.max_samples))
+    max_samples = config.evaluation.max_samples
+    if max_samples is not None and max_samples > 0:
```
veekaybee, line 94:
```suggestion
    if max_samples and max_samples > 0:
```
we can simplify to check for truthiness here, i.e. checking for both not None and not 0

src/lm_buddy/jobs/evaluation/hf_evaluate.py
```diff
@@ -99,14 +104,18 @@ def run_eval(config: HuggingFaceEvalJobConfig) -> Path:
     # Choose which model client to use
     if type(config.model) == VLLMCompletionsConfig:
         model_name = config.model.inference.base_url
+        output_model_name = config.model.inference.engine
+        if "mistral" in model_name:
+            # run the mistral client
+            logger.info(f"Using Mistral client. Endpoint: {model_name}")
+            model_client = MistralModelClient(model_name, config.model)
+        else:
```
veekaybee, line 112: could there be a case that the model is not mistral and not openAI and we try to also run the OpenAI client on it? we might want an elif here if that's the case

src/lm_buddy/jobs/model_clients.py
```diff
@@ -82,7 +84,52 @@ def predict(self, prompt):
...
+
+    def __init__(self, config: VLLMCompletionsConfig):
+        self._config = config
+
+        hf_model_loader = HuggingFaceModelLoader()
+        self._engine = hf_model_loader.resolve_asset_path(config.inference.engine)
+        self._system = config.inference.system_prompt
+
+    @abstractmethod
+    def _chat_completion(
```
veekaybee, line 98: since an abstract method doesn't usually contain implementation, we might want to take out the system prompt, or alternatively make it a non-abstract method

src/lm_buddy/jobs/evaluation/hf_evaluate.py
aittalam: Not right now, i.e. all the other models (VLLM, llamafile) while not beign OAI use the OAI client (because they wanted to be compatible). But I agree we'll need to have elifs when we start using other clients (or, better, explicitly write a mapping between API URLs and Clients + fallback on OAI for generic IP addresses)

src/lm_buddy/jobs/model_clients.py
aittalam: +100 I removed the default value as it did not make a lot of sense anyway

veekaybee (approved):
LGTM once comments are addressed, 🔥

aittalam:
> LGTM once comments are addressed, 🔥

Thank you!
The lm-harness entrypoint still fails tests but we can ignore it for now (it works when running integration tests locally, there might be a library version mismatch to fix)

PR #113, opened 2024-08-05
Added model_max_length as a tok parameter

aittalam:
Since 4.40.0, transformers does not get model_max_tokens from the model's family dict but explicitly looks for this parameter in `tokenizer_config.json`. Not all models provide this, so one possibility is to explicitly allow users to provide this parameter in the config. This PR implements exactly this.

Tested with facebook/bart-large-cnn + pytest unit/integration tests.

### DIFF
--- a/src/lm_buddy/jobs/model_clients.py
+++ b/src/lm_buddy/jobs/model_clients.py
@@ -44,10 +44,16 @@ class SummarizationPipelineModelClient(BaseModelClient):
     (model is loaded locally).
     """
 
-    def __init__(self, model: str, config: AutoModelConfig):
+    def __init__(self, model: str, config: HuggingFaceEvalJobConfig):
+        self._config = config
+
+        hf_tokenizer_loader = HuggingFaceTokenizerLoader()
+        self._tokenizer = hf_tokenizer_loader.load_pretrained_tokenizer(config.tokenizer)
+
         self._summarizer = pipeline(
             "summarization",
             model=model,
+            tokenizer=self._tokenizer,
             device=0 if torch.cuda.is_available() else -1,
             truncation=True,
         )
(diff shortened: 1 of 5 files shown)


src/lm_buddy/jobs/model_clients.py
```diff
@@ -44,10 +43,16 @@ class SummarizationPipelineModelClient(BaseModelClient):
     (model is loaded locally).
     """
 
-    def __init__(self, model: str, config: AutoModelConfig):
+    def __init__(self, model: str, config: HuggingFaceEvalJobConfig):
```
binaryaaron, line 46: Double check the type here here?

binaryaaron (approved):
approving given the internal constraints; double check a few things before landing though. I'd suggest adding an extra test to the unit tests to check for it.

src/lm_buddy/jobs/model_clients.py
aittalam: Thanks for the heads up! Yep, the type is right: before I was not using the config as the pipeline just needed the model name, so I had decided to provide only the simplest config possible. Now I am passing the same config as with general HF models as I am gonna instantiate a tokenizer with that (it's even better as it makes the two calls consistent with each other)

```
        # depending on config, use the summarizer pipeline or directly call the model
        # for inference
        if config.evaluation.use_pipeline:
            logger.info(f"Using summarization pipeline. Model: {model_name}")
            model_client = SummarizationPipelineModelClient(model_name, config)
        else:
            logger.info(f"Using direct HF model invocation. Model: {model_name}")
            model_client = HuggingFaceModelClient(model_name, config)
```

aittalam:
> approving given the internal constraints; double check a few things before landing though. I'd suggest adding an extra test to the unit tests to check for it.

Thanks! I have two more PRs in queue:
- bump of all libs (this will allow us, anmong other things, to run more recent models)
- adding text generation pipeline + prompts

Will make unit tests for both and add one for this change too (explicitly passing a longer text as input and verifying it does not break with the new setup)

PR #114, opened 2024-08-06
Fixes flakey integration test/syncs deps

aittalam:
## What's changing

I updated pyproject.toml to use more recent libs. The main reason was transformers so we could run more recent models but I ran a general update using `pip-review` to get latest versions. These are the steps I followed:

- used pip-review to update all libs
- check there were no inconsistencies
- test lm-buddy with the new set of libs
- updated `pyproject.toml` by picking the updated versions from the output of `pip freeze`
- re-installed everything in a new env and made sure there were no inconsistencies
- tested lm-buddy again

## How to test it

The tests I ran were unit/integration tests + manually running lm-buddy for summarization eval (currently our main job).
Interestingly, `tests/integration/test_lm_harness.py::test_lm_harness_job` passes when I test locally

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,47 +26,45 @@ dependencies = [
     "wandb>=0.16.3",
     "protobuf>=3.20.2",
     "urllib3>=1.26.18,<2",
-    "pydantic>=2.6.4",
-    "pydantic-yaml>=1.2.0",
     "pydantic==2.6.4",
     "pydantic-yaml==1.2.0",
     "ray[default]==2.30.0",
     "loguru==0.7.2",
-    "s3fs",
+    "s3fs==2024.5.0",
 ]
 
 [project.optional-dependencies]
-ruff = ["ruff==0.2.1"]
+ruff = ["ruff==0.5.6"]
 jobs = [
     # HuggingFace / pytorch
-    "torch==2.3.1",
-    "datasets>=2.17.1",
-    "transformers==4.38.0",
-    "accelerate==0.26.1",
-    "peft==0.7.1",
-    "trl==0.7.10",
+    "torch==2.4.0",
+    "datasets>=2.20.0",
+    "transformers==4.43.4",
+    "accelerate==0.33.0",
+    "peft==0.12.0",
+    "trl==0.9.6",
     "bitsandbytes==0.42.0",
     "bert_score==0.3.13",
     # Evaluation frameworks
-    "lm-eval==0.4.2",
-    "einops==0.7.0",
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -26,47 +26,45 @@ dependencies = [
     "wandb>=0.16.3",
     "protobuf>=3.20.2",
     "urllib3>=1.26.18,<2",
-    "pydantic>=2.6.4",
-    "pydantic-yaml>=1.2.0",
```
aittalam, line 30: I kept only the == as there was redundancy here and having both would have resulted in having ==, but am happy to put >= if the intention was to keep them updated

pyproject.toml
```diff
@@ -26,47 +26,45 @@ dependencies = [
...
     "pydantic==2.6.4",
     "pydantic-yaml==1.2.0",
     "ray[default]==2.30.0",
     "loguru==0.7.2",
-    "s3fs",
+    "s3fs==2024.5.0",
```
aittalam, line 33: pinned to what we are actually running rn

pyproject.toml
```diff
@@ -26,47 +26,45 @@ dependencies = [
...
 ]
 
 [project.optional-dependencies]
-ruff = ["ruff==0.2.1"]
+ruff = ["ruff==0.5.6"]
```
aittalam, line 37: @binaryaaron is this ok? Not sure whether this has to be changed anywhere else for testing on github

aittalam:
Two questions for @binaryaaron: 

1) ruff on github returns an error which I do not get at precommit nor when I run it manually. Is there anything else I should change in the repo to make sure the two are aligned? (nothing major, I'd just like to prevent errors happening here if I can)

2) I still wonder what is different between my local tests and github's as they all pass locally (in particular the lm-harness one which breaks online):
image

pyproject.toml
binaryaaron: yeah, i changed it back to make sure it worked via pip-compile.

pyproject.toml
binaryaaron: made it match what's being used internally for now

binaryaaron:
I added some minor changes and changed the test task to one that doesn't require HF remote code. there's something odd with how those args are being passed; i opened #115 to revisit this after the demo.

binaryaaron (approved):

aittalam:
> I added some minor changes and changed the test task to one that doesn't require HF remote code. there's something odd with how those args are being passed; i opened #115 to revisit this after the demo.

That's great, many thanks!

PR #119, opened 2024-09-16
Hard pin datasets and wandb

veekaybee:
## What's changing

We're pinning wandb and datasets due to regressions from upper-bound versions because integration tests were failing with updated versions. 

## How to test it

Run `pytest -s tests/integration/test_finetuning.py` locally


binaryaaron (approved):
lgtm 😅 👋🏽 😛

veekaybee:
❤️

chainlink (approved):

PR #121, opened 2024-09-19
Notice of archive of repo 

veekaybee:
## What's changing
 lm-buddy is [being folded into Lumigator](https://github.com/mozilla-ai/lumigator/pull/209). We'd like to make sure people understand the change before we archive the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,9 @@
 # LM Buddy
 
-> [!WARNING]
+> [!IMPORTANT]
 >
-> LM Buddy is in the early stages of development.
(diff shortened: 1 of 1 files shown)


aittalam (approved):
LGTM, thanks Vicki!
I will add an issue in Lumigator to provide documentation [here](https://github.com/mozilla-ai/lumigator/tree/main/lumigator/python/mzai/lm_buddy) on how lm-buddy is ran as a command-line job, so ppl won't have to look into the code to see how it works

